Issue #14, opened 2019-07-14
Stop using openapi-* dependencies 


sheldhur:
Why? 

`openapi-request-coercer` - `Ajv` can do it
`openapi-default-setter` - `Ajv` can do it
`openapi-request-validator` -  it's wrapper over `Ajv` that can't correctly resolve $ref. But `Ajv` no have troubles with resolving $ref

cdimascio:
@sheldhur thanks for the recommendation. this is a great idea to remove the dep on openapi-request-validator. I will look into this. 
please let me know, if already have some suggestions. if so, please post them here or create a PR.
thanks so much for helping to improve this package!

sheldhur:
@cdimascio I made this implementation, you can take this as a basis
https://gist.github.com/sheldhur/5111f3afee51de6698dfc626d65f95c5

cdimascio:
@sheldhur much appreciated! i will have a look at this and work to remove the dependency. if you have something working feel free to submit a PR. i'll gladly merge if it looks good. 
again, thanks for your help!

cdimascio:
@sheldhur, i've added much of your validator gist into the branch `req_validator`. It's not working yet and most tests fail. In any case, I did want to at least get the bulk of your code in there, so that I can experiment with it. Feel free to contribute as well.

cdimascio:
diff is [here](https://github.com/cdimascio/express-openapi-validator/compare/req_validator?expand=1)

cdimascio:
made a bit of progress this evening WITH @sheldhur's validator

```
  28 passing (292ms)
  18 failing
```

sheldhur:
@cdimascio my code support only `application/json` content-type in request body. You can try change 
`schema.content[TYPE_JSON].schema` 
on 
`schema.content[TYPE_JSON].schema || schema.content[TYPE_FORM].schema` 
in initAjv for support `multipart/form-data`. But it not will work with several content types request body schema. If it is critical, I will try to fix it on the weekend

https://github.com/cdimascio/express-openapi-validator/blob/e934e372e9aab4deafaa2d3223cd1cb161848e44/src/middlewares/openapi.request.validator.2.ts#L97

cdimascio:
That would be fantastic

cdimascio:
@sheldhur  There appears to be an issue with request bodies. 

For example, if you run the test below you'll see that the body is not being validated by Ajv. 
```
      it.only('should return 400 if required "name" property is missing', async () =>
[...]

PR #17, opened 2019-07-20
Parameter deserialization support

sheldhur:
* add tests for JSON in params
* add deserialization separated array in params + tests

#14


cdimascio:
merged. thank you!

PR #38, opened 2019-08-19
Add support for using $ref in requestBody

trebler:
This PR adds support for referencing ```requestBody``` from ```components/requestBodies```


trebler:
@sheldhur @cdimascio

Issue #45, opened 2019-08-28
Reject requests with null value if "nullable" is not set


tafm:
Actually its the same with or not the nullable field

cdimascio:
changing this from bug to enhancement as it hasn't been implemented yet.

cdimascio:
@tafm I've started some work in this area
please see https://github.com/cdimascio/express-openapi-validator/pull/46 for details

this set the foundation to solve the issue you described.
ultimately, it makes express-openapi-validator aware of `nullable`. essentially, if `nullable: true` it will allow the null value to pass through. if `nullable: false` or nullable is not provided, it will convert the value to its `default` value. in the case of string, empty string.

cdimascio:
@sheldhur do you know whether its possible to turn off defaults for required parameters

doing so will be one way to solve this using v1.7.0
also, it aligns with avoiding [Common Mistakes](https://swagger.io/docs/specification/describing-parameters/) (as described in the OpenAPI spec)
> **Common Mistakes**
There are two common mistakes when using the default keyword:
Using default with required parameters or properties, for example, with path parameters. This does not make sense – if a value is required, the client must always send it, and the default value is never used.

sheldhur:
@cdimascio hi. You can change keyword behavior. 

https://ajv.js.org/custom.html

cdimascio:
@tafm You can now get the desired behavior by setting `coerceTypes: false`. This will ensure that `null` values are not coerced to the default value of `type`specified in the spec. e.g. `type: string` would coerce a `null` to `''`

v1.8.0 allows you to set coerceTypes to false

tafm:
@cdimascio thank you very much, this is exactly i need to enable in production. Im still unnable to use now, but not really for a "lack" of feature but some incompatibility between json schemas and open api (setting coerceTypes to false makes ajv invalidate: "not of type string"). I had a look at the code and how to fix, one way its using the library **openapi-schema-to-json-schema**

`const toJSONschema = require('openapi-schema-to-json-schema')`
then replace `const validator = this.ajv.compile(schema);` to `const validator = this.ajv.compile(toJSONschema(schema));` this converts types with `nullable: true` for a array of types `[oldType, null]`. Strangelly i think a better idea would be replace to anyOf, but works.

Another little issue its that `coerceTypes` is hardcoded to true at the npm repository / js files, the .ts files are right on github.

cdimascio:
[...]

PR #46, opened 2019-08-30
Reject requests with null value if "nullable" is not set #45

cdimascio:


cdimascio:
This provides support for nullable values, however it does not reject requests if null is set.

The behavior is as follows:

**Case 1**
nullable is set to true on field name:
```yaml
    NullableNewPet:
      additionalProperties: false
      required:
        - name
      properties:
        name:
          type: string
          nullable: true
        tag:
          type: string
```
This allows a caller to pass null and the value will not be converted to a default value of e.g. empty string. Instead. the null will be preserved

**Case 2**

nullable: false i.e. the default case
```yaml
    NewPet:
      additionalProperties: false
      required:
        - name
      properties:
        name:
          type: string
        tag:
          type: string
```

If a caller passes null, name is default to empty string (or a `default` value if specified)

Currently, to reject a null, must be handled by the implementor.

As such we I will keep this ticket open.

cdimascio:
solves Make express-openapi-validatator aware of the OpenAPI `nullable` property #47
and related to #45

PR #51, opened 2019-09-15
Response validation

cdimascio:


cdimascio:
@sheldhur any comments/suggestions on this as a starting point for response validation?

Note that I instantiate a second Ajv instance to be used for response validation.

That second instance, it is almost identical to request validation Ajv instance, with the exception that it does not modify the `readOnly` keyword

Here is a pointer to that conditional logic
https://github.com/cdimascio/express-openapi-validator/pull/51/files#diff-248a764b6bcd842ea00092eb738952b8R30

Would love to know if believe there are additional changes that should be incorporated

Thanks!

PR #54, opened 2019-09-22
Document how validation only applies underneath the server's base URL

### LINKED ISSUE #52: do not validate request if path does not exist in openapi spec
I am serving frontend and restful APIs within a same app. I want to validate only my restful APIs which have a openapi spec, however, my frontend pages are throwing not found errors because these paths do not exist in the specification.

Is there a way to limit validation to a certain path, for example, /api/? Or even disable not found errors for paths not found in openapi spec?

richdouglasevans:
closes #52

Screen Shot 2019-09-22 at 11 28 12

The README is getting a bit long now and this is just bodged in there: it's not especially discoverable. This section doesn't say what to do if you're co-locating your frontend and API but under the same base URL: i.e. what do you do when you're _not_ using something like `/api` or `/v1` in your base URL? I'd like to expand on that but now we're veering into documentation as _guide_ rather than documentation as _reference_ so I've chosen to keep it terse.

I've submitted this PR anyway 'cos the README can be structured and hyperlinked to form better reference documentation in a subsequent PR, yeah?

### DIFF
--- a/README.md
+++ b/README.md
@@ -159,6 +159,27 @@ app.use((err, req, res, next) => {
 });
 ```
 
+## The Base URL
+
+The validator will only validate requests — and (optionally) responses — that are under
+the server's [base URL](https://spec.openapis.org/oas/v3.0.0.html#serverVariableObject).
+
+This is useful for those times when the API and frontend are being served by the same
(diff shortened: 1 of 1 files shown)


cdimascio:
Looks good. thank you!

cdimascio:
@all-contributors please add @richdouglasevans for documentation

Issue #34, opened 2019-08-17
Add support for JSONSchema readOnly keyword


cdimascio:
Support readonly keyword 
https://json-schema.org/latest/json-schema-validation.html#rfc.section.10.3
Note: Ajv supports readOnly

cdimascio:
reposting the following info provided by @gonenduk

For example, when I create a new product - I supply several fields and the product is created in the db and an id is given to it, as well as ownerId and creation time. The newly created object is returned with the new values - all of them.
Later on I can use this id to retrieve the product, delete it or update it. But I can never change the id or supply it when creating a new product (nor the creation time or the ownerId).
Read-only fields supplied on post or put or patch should not pass validation.
```
components:
  schemas:
    Product:
      type: object
      properties:
        _id:
          type: string
          readOnly: true
        title:
          type: string
        price:
          type: number
        ownerId:
          type: string
          readOnly: true
        createdAt:
          type: string
          format: date-time
          readOnly: true
        updatedAt:
          type: string
          format: date-time
          readOnly: true
```
The read-only fields are returned on get (and maybe delete if you returned the deleted object), but cannot be given on post, put and patch!

cdimascio:
https://swagger.io/docs/specification/data-models/data-types/

gonenduk:
And more specific here: https://swagger.io/docs/specification/data-models/data-types/#readonly-writeonly

cdimascio:
For those participating in Hacktoberfest who want to look into this issue, See the PR above for getting started tips. Some implementation hints are provided

cdimascio:
v2.8.0 provides support for `readOnly` properties. thanks @gonenduk @frankcalise!

PR #71, opened 2019-10-14
Add support for JSONSchema readOnly keyword

cdimascio:
I've provided this PRs as help for folks participating in #hacktoberfest who may want to pick up issue #34

This PR creates a basic test for `readOnly` property support.

In order to solve this, modifications must be made to `validate` in [openapi.request.validator.ts](https://github.com/cdimascio/express-openapi-validator/blob/master/src/middlewares/openapi.request.validator.ts#L22).

One possible implementation may be to check the schema e.g. `req.openapi.schema` for properties marked readonly, if found in the user request e.g. `req.body` throw an error.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -1,0 +1,22 @@
+# contributing to express-openapi-vaidator
+
+## Prerequisites / Setup
+
+- Fork the repo
+- Clone your copy `git clone <forked-repo>`
+- `npm install`
+
+## Run the tests
+
+```shell
+npm test
+```
+
+## Project structure
+
+`src` contains the source code
+`test` contains the tests
+
+## Need help
+
+Reach out on [gitter](https://gitter.im/cdimascio-oss/community)
--- a/README.md
+++ b/README.md
@@ -159,7 +159,7 @@ Determines whether the validator should coerce value types to match the type def
 
 ### securityHandlers (optional)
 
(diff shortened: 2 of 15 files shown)


cdimascio:
@frankcalise Here are a few more hints

on line 61, you will have the `requestBody` sent by the caller
The requestBody must be inspected for `readOnly` properties
To find which props are readonly, you need to look at the schema definition
(The test uses a $ref)
To get the schema from a $ref you can do this

```javascript
const r = this.ajv.getSchema(requestBody.content['application/json'].schema['$ref'])`
console.log(r.schema)
```

This will output
```json
"{
  "type": "object",
  "properties": {
    "id": {
      "type": "string",
      "readOnly": true
    },
    "name": {
      "type": "string"
    },
    "price": {
      "type": "number"
    },
    "created_at": {
      "type": "string",
      "format": "date-time",
      "readOnly": true
    }
  }
}"
```

Here you can see which properties are `readOnly`. If the requestBody contains `readOnly` property and error should be thrown

cdimascio:
i've reworked this a bit to provide a custom keyword via Ajv. This is more efficient as it does not require recursively express-openapi-validator to recursively resolve the schema against the user's request payload. Ajv already makes a pass for validation, hence we can piggyback onto that and perform our checks during the Ajv validation pass.

PR #75, opened 2019-10-18
Fix #74

### LINKED ISSUE #74: Can't use multiple Media Types in combination with "multipart/form-data" in `requestBody.content`
I am trying to create a route schema, which supports two requestBody contents (`application/json` and `multipart/form-data`). When I send `application/json` request, `unsupported media type` error is returned. `multipart/form-data` works ok.

I think the problem is in `isMultipart(req)` function, which does not predict that other content types might be present in schema configuration. (https://github.com/cdimascio/express-openapi-validator/blob/522c057ca4a16c06cb97f2f681b93f4d425ccbdc/src/middlewares/openapi.multipart.ts#L48) 

I think that `req.headers['content-type']` should also be checked together with `isMultipart(req)`.
https://github.com/cdimascio/express-openapi-validator/blob/522c057ca4a16c06cb97f2f681b93f4d425ccbdc/src/middlewares/openapi.multipart.ts#L8

Maybe this simple solution would solve this problem :)
```js
if (isMultipart(req) && isValidContentType(req)) {
      mult.any().......
}
```
In case of invalidContentType, unsupported media type error will be thrown later in code flow.

medolino:
I implemented proposed fix from issue https://github.com/cdimascio/express-openapi-validator/issues/74

### DIFF
--- a/src/middlewares/openapi.multipart.ts
+++ b/src/middlewares/openapi.multipart.ts
@@ -5,10 +5,7 @@ import * as multer from 'multer';
 export function multipart(openApiContext: OpenApiContext, multerOpts: {} = {}) {
   const mult = multer(multerOpts);
   return (req, res, next) => {
-    if (isMultipart(req)) {
-      if (!isValidContentType(req)) {
-        throw validationError(415, req.path, 'unsupported media type');
-      }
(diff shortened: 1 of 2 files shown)


cdimascio:
@all-contributors please add @medolino for code

Issue #99, opened 2019-10-29
Strict contract responses


plastic-abubakr:
Is there a configuration that I'm missing that strips out any fields that are not defined in the response schema?

cdimascio:
thanks for the issue @plastic-abubakr.
currently, we do not strip out fields not defined in the response schema, however its certainly worth consideration.
can you describe your use case, the desired behavior, and an example. thanks!

cdimascio:
prior to a recent fix, responses were allowing additional properties to pass validation. we could consider making this option again. is that what you are looking for?

cdimascio:
@plastic-abubakr ^

cdimascio:
Got it. This makes sense. I'll add support for this.

richdouglasevans:
By all means add support for this: I understand the issue and the motivation 👍 

This does expand the scope of this library though, a library which has `validator` in the project name. My expectation of a validator is that it will _validate_: if you add support for transformation also the library will continue to validate… but, I don't expect a validator to _transform_.

Transformation of either the input or the output to conform to some schema is a distinct concern. Even if the transformation feature is optional — and I expect it'll be off by default — adding this feature dilutes the focus of this library. It'd be helpful for some set of users, undoubtedly, but it also adds weight that's not directly relevant to what this library is about: validation.

Perhaps there's a distinct library that could be developed — `express-openapi-transformer` — that does transformation only? An application might compose these to achieve the effect of both validation and transformation.

Like I say, do add support for this if you're so minded 👍 😃

comino:
@richdouglasevans I tend to disagree. The widely used JSON validation library "joi" also decided to support "stripUnknown" to strip unknown fields from a payload. Users like me would end up in adding Joi just to implement a sanitizer to request and response. 

Personally I would like to see this feature ( but would disable it by default).

cdimascio:
in v2.14.2 you can get the behavior you are looking for with the following option:

```
validateResponses: {
   removeAdditional: 'failing'
 }
```

This will validate responses, but remove any additionalProperties that would not pass schema validation.

plastic-abubakr:
This does not seem to be working

> in v2.14.1 you can get the behavior you are looking for with the following option:
> 
> ```
> validateResponses: {
>    removeAdditional: 'failing'
>  }
> ```
> 
> This will validate responses, but remove any additionalProperties that would not pass schema validation.

cdimascio:
[...]

PR #101, opened 2019-11-02
#99 option to remove additional props from responses

cdimascio:


src/index.ts
```diff
@@ -42,6 +45,18 @@ export class OpenApiValidator {
...
     if (options.coerceTypes == null) options.coerceTypes = true;
     if (options.validateRequests == null) options.validateRequests = true;
+    if (options.validateResponses == null) options.validateResponses = false;
+    if (!options.validateRequests) throw Error('validateRequests must be true');
+
+    if (!options.validateResponses) {
+    } else if (
+      options.validateResponses === true ||
+      options.validateResponses === 'strict'
+    ) {
```
richdouglasevans: The empty `if` block can be elided.

```ts
if (
    options.validateResponses === true ||
    options.validateResponses === 'strict'
  ) {
    options.validateResponses = {
        removeAdditional: false,
    };
}
```

Also, [the `validateResponses` docs](https://github.com/cdimascio/express-openapi-validator#%EF%B8%8F-validateresponses-optional) state that the value must be a `boolean`. There is no mention of the string literal `"strict"` and this literal doesn't appear in the tests.

So, can this be further boiled down to:

```ts
if (options.validateResponses === true) {
    options.validateResponses = {
        removeAdditional: false,
    };
}
```

src/index.ts
```diff
@@ -42,6 +45,18 @@ export class OpenApiValidator {
     if (options.unknownFormats == null) options.unknownFormats === true;
     if (options.coerceTypes == null) options.coerceTypes = true;
     if (options.validateRequests == null) options.validateRequests = true;
+    if (options.validateResponses == null) options.validateResponses = false;
+    if (!options.validateRequests) throw Error('validateRequests must be true');
```
richdouglasevans: 🤔

Does this assertion also need to be reflected in the documentation? Currently [the `validateRequests` docs](https://github.com/cdimascio/express-openapi-validator#%EF%B8%8F-validaterequests-optional) state that the parameter is optional and defaults to `true`… but if I set it to `false` then the app is going to error.

Perhaps the parameter needs to be marked as deprecated or the documentation changed to reflect that the value  is still optional but *must* be `true`.

src/index.ts
```diff
@@ -95,11 +110,15 @@ export class OpenApiValidator {
     const requestValidatorMw: OpenApiRequestHandler = (req, res, next) =>
       requestValidator.validate(req, res, next);
 
+    const removeAdditional =
+      this.options.validateResponses &&
+      (<ValidateResponseOpts>this.options.validateResponses).removeAdditional;
```
richdouglasevans: `Minor`

Sure, this works but after construction the `this.options.validateResponses` is always a non-null `ValidateResponseOpts` instance. This isn't captured in the types so that's why this code does this awkward "check for null and then cast" dance.

I just wanna type this:

```ts
const removeAdditional = this.options.validateResponses.removeAdditional;
```

Re-using the same type for the initialisation parameters _and_ the resolved options is the root cause here. It's only one occurrence of this awkward dance so we could just let it go but it could be the thin end of the wedge and, as more options are added in future, we could see this kind of code littering this class.

If you agree that this is an issue — and hey, such design stuff is often subjective — then I suggest raising a ticket to address this in a distinct PR: that way this PR and feature can progress and the internal design issue around typing can be addressed separately.

src/index.ts
cdimascio: good point. I will update the doc. technically it is optional, but there are no tests for disabling request validation. i also wonder if there is value there. all in all, i like the idea of deprecating it in 2.x and removing it in the next major release. 
i'll make that update

src/index.ts
cdimascio: thanks again @richdouglasevans. i'll make this change as well

src/index.ts
cdimascio: modified the PR to remove 'strict'

src/index.ts
cdimascio: validateRequests: false is allowed again, but deprecated

src/index.ts
cdimascio: cleaned up middleware registration and removed the 'ugly' assignment

src/index.ts
cdimascio: err. im waffling a bit on keeping `validateRequests` optional, rather than required. I'll keep it as supported for now...

cdimascio:
Thanks for all the feedback, @richdouglasevans!

Issue #108, opened 2019-11-06
incorrectly erroring on nullable date-time


SpencerLawrenceBrown:
Hi,

I think there is an issue with how nullable date-time fields are handled.

When a property is `type: string, format: date-time, nullable: true`, the null value is coerced into an empty string which fails the date-time validation.

I know this because if I remove the `format: date-time` in my example then 
```
{
     completedAt: null
}
```
becomes
```
{
    completedAt: ''
}
```
and passes validation.

Example that breaks:

`yaml`
```
WorkflowStep:
      properties:
        completedAt:
          description: If it hasnt been completed, value will be null.
          type: string
          nullable: true
          format: date-time
      required:
        - completedAt
```

`object`
```
{
    completedAt: null
 }
```

`error`
```
{ path: 'workflowStep.completedAt',
  message: 'should match format "date-time"',
  errorCode: 'format.openapi.validation' },
```

SpencerLawrenceBrown:
I think I found the issue and I will open a PR

cdimascio:
Thanks @SpencerLawrenceBrown. the fix is available in 2.14.2

PR #109, opened 2019-11-06
add null value support to response validation

SpencerLawrenceBrown:
Addressing the issues mentioned in: https://github.com/cdimascio/express-openapi-validator/issues/108

Null values were not valid in response objects. Instead values were being coerced to default for type according to AJV. So strings became `''` and numbers became `0`.

Added `nullable: true` to the options passed to AJV. This mimics the behavior of the request validator.

Also cleaned up an if statement checking equality using `=` instead of `==`.


cdimascio:
@all-contributors add @SpencerLawrenceBrown for code and test

Issue #113, opened 2019-11-08
Support custom basePath


henhal:
I have an OpenAPI 3.0.2 definition which I use to create the public documentation for my API. In the spec, I list the URL:s of my public servers under the `servers` section.
Now I am trying to use the same definition document with express-openapi-validator to validate my requests, but my backend server is running in a private VPC behind a proxy and is not using the same base path as the public facing API served through AWS API gateway.

To solve this it would be great to be able to supply a custom basePath to the options passed to the OpenApiValidator, instead of having it parse the `servers` section.

Example:

document:
```
openapi: 3.0.2
info:
  title: foo
servers:
  - url: 'https://my-public-api.com/some/path/prefix'
```

API gateway will now route `/some/path/prefix/*` to my Express application, without the prefix.
When I attach the OpenApi validator, it will however look for route `/foo` in the open api route map as `/some/path/prefix/foo`, which does not exist.

It seems this could be solved by allowing a custom basePath to be supplied through the constructor options, which would then mean ignoring the `servers` section?

henhal:
Assuming PR's are welcome, I provided one for you to look at. 

Thanks.

cdimascio:
@henhal PRs are always welcome. We love contributors! Thanks for you interest in the project :)

With respect to the servers proposal... 
I'm reluctant to add this feature for two reasons:
1. The performance (despite doubling the route count) is negligible. For example, on my Macbook, I can perform well over 200,000 lookups per millisecond on a js object containing 100,000 entries - see https://jsperf.com/javascript-objects-vs-map-performance. All in all, the performance aspect is insignificant. Also, the number of routes in an api spec, even when doubled, triple, quadrupled, ... will certainly not approach 100k.

2. Given that performance is effectively a non-issue, I'd prefer not to add such additional options as it clutters the usage options and expands scop.

That all being said and assuming you still do **not** want to include both base paths in the openapi spec when running in production, I would recommend making this modification during deployment step. For example, you might a yaml parser e.g. `js-yaml` or `yq` to automate the removal the of dev server entry when deploying to prod, and vice versa for dev. There are many tools and optoins available to do this.

Hope this helps!

henhal:
Thanks for the input. I honestly don't think a base path option is very cluttering, and other open api adapters I've worked with provide this, but I see your point. I actually already use a preprocessing step on my spec (with yimp) so I guess I could solve it there though.

[...]

PR #115, opened 2019-11-08
Add basePath option to support custom basePath over server base URL

henhal:
Adding an option basePath used to customize the base path on which all routes are mounted.
By default, the base path is derived from the server(s) defined in the OpenAPI specification, but this option may be used if the backend server implementing the API is mounting the routes on another path.


henhal:
Fixes #113

xg1990:
Hi @henhal , is there a workaround for this issue. We are also facing this problem  😭

Issue #119, opened 2019-11-11
Validator is caching wrong API descriptor


rafalneves:
If you have express routes layout like this:

_express.js:_
` 
app.use('/service', authRouter);
`
_authRouter.js_
`
serviceRouter.route('/alive')
  .get((req, res) => {
    res.sendStatus(200);
  });
`

Response req.path = /alive

In my case I was Mocha testing two consecutive routes /service/alive and /app/alive (examples) and getting errors in the second because was being validated with the firsts API description. Was going crazy...

So, openapi.response.validator.ts **49** `key = ${req.method}-${req.path}-${contentType} ` does not guarantee uniqueness.

cdimascio:
Thanks for the issues and the fix @rafalneves. Looking at the PR now

cdimascio:
fix merged and available in `v2.14.5`

PR #120, opened 2019-11-11
Fix req.path uniqueness in response validators caching.

rafalneves:
#119


cdimascio:
@rafalneves given your find, `originalUrl` should also be used in the cache key for requests. i've added the change to your PR

cdimascio:
@all-contributors add @rafalneves for code

PR #128, opened 2019-11-16
Fix #124

### LINKED ISSUE #124: unsupported media type for "application/json; charset=utf-8"
Hi, my company has a rule that we mandatory set `Content-Type: application.json; charset=utf-8` to all the request/response header. (I know the `charset=utf-8` is probably redundant for json though.)
Since OpenAPI Validator drop charactors of Content-Type after the charactor of `;` as below, our OpenAPI Validator cannot find the appropriate request body schema and respond `unsupported media type application/json`.

https://github.com/cdimascio/express-openapi-validator/blob/04a84e513ab253f7d7e655c53bdac7978ba43093/src/middlewares/util.ts#L6-L17

Sample api spec is described as below. Due to the issue above, requests which has `Content-Type: application.json; charset=utf-8` cannot be associated with the schema `$ref: '#/components/schemas/user'`.

```yaml
paths:
  /user/:userId:
    post:
# --- skip ---
      requestBody:
        content:
          application/json; charset=utf-8:
            schema:
              $ref: '#/components/schemas/user'
# --- skip ---
```

I think we should use original raw value of `Content-Type` in request header to find appropriate schema.

https://github.com/cdimascio/express-openapi-validator/blob/04a84e513ab253f7d7e655c53bdac7978ba43093/src/middlewares/openapi.request.validator.ts#L196 

If you feel okay with my suggestion, let me create PR.

mk811:
**[Objective]**
To allow `charset` in content-type header.

**[Related Issue]**
#124 

**[Note]**
Since the `boundary` will be automatically filled, we should drop as before.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -2457,9 +2457,9 @@
       "dev": true
     },
     "handlebars": {
-      "version": "4.5.1",
-      "resolved": "https://registry.npmjs.org/handlebars/-/handlebars-4.5.1.tgz",
-      "integrity": "sha512-C29UoFzHe9yM61lOsIlCE5/mQVGrnIOrOq7maQl76L7tYPCgC1og0Ajt6uWnX4ZTxBPnjw+CUvawphwCfJgUnA==",
+      "version": "4.5.2",
+      "resolved": "https://registry.npmjs.org/handlebars/-/handlebars-4.5.2.tgz",
+      "integrity": "sha512-29Zxv/cynYB7mkT1rVWQnV7mGX6v7H/miQ6dbEpYTKq5eJBN7PsRB+ViYJlcT6JINTSu4dVB9kOqEun78h6Exg==",
       "dev": true,
       "requires": {
         "neo-async": "^2.6.0",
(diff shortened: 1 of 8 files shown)


cdimascio:
@mk811 thanks for the PR. I may a few changes to ensure backward compatibility as well as a means to map equivalent content types. note these updates will enable both `application/json` and `application/json; charset=utf-8` to pass validation when openapi spec defines one or the other. i believe this makes sense given they are semantically equivalent. thoughts?

mk811:
@cdimascio 
I like your idea to treat these equivalently. I'm afraid that current program ensures compatibility for `application/json` only.
To cover any other content types such as `application/xml`, I added updates a little.

The risk of adding `charset=utf-8` to `mediaType` at expres-validator is not considerable, since `this.withoutBoundary` will be evaluated first.

```javascript
  equivalents(): string[] {
    if (!this.withoutBoundary) return [];
    if (this.charSet) {
      return [this.mediaType, `${this.mediaType}; ${this.charSet}`];
    }
    return [this.withoutBoundary, `${this.mediaType}; charset=utf-8`];
  }
```

cdimascio:
@all-contributors add @mk811 for code and test

PR #138, opened 2019-11-24
validate if server object is empty in an openapi definition.

### LINKED ISSUE #137: Validator does not work with OpenAPI v3 definition with empty servers object.
This issue can be reproduced with the given example on repo, just switching
https://github.com/cdimascio/express-openapi-validator/blob/master/example/example.yaml#L10
```yaml
servers:
  - url: http://petstore.swagger.io/v1
```
for
```yaml
servers: []
```

endpoints should be changes on https://github.com/cdimascio/express-openapi-validator/blob/master/example/app.js#L36-L57 from 
```
'/v1/pets'
```
to 
```
'/pets'
```
other way, we would get a `404` `v1/pets` not found.

So having that, requesting to example server using the given CURL samples on page https://github.com/cdimascio/express-openapi-validator/tree/master/example#validate-a-query-parameter-with-a-value-constraint
will not work and endpoints will be reached.

The reason of this behavior is here:
https://github.com/cdimascio/express-openapi-validator/blob/master/src/framework/util.ts#L54

So, changing
```js
if (!servers) {
    return [new BasePath({ url: '' })];
 }
```
to
```js
if (!servers || servers.length == 0) {
    return [new BasePath({ url: '' })];
 }
```
will solve the issue

HugoMario:
fixes #137

### DIFF
--- a/src/framework/util.ts
+++ b/src/framework/util.ts
@@ -51,7 +51,7 @@ export function sortApiDocTags(apiDoc) {
 export function getBasePathsFromServers(
   servers: OpenAPIV3.ServerObject[],
 ): BasePath[] {
-  if (!servers) {
+  if (!servers || servers.length == 0) {
     return [new BasePath({ url: '' })];
   }
   const basePathsMap: { [key: string]: BasePath } = {};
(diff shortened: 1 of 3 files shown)


cdimascio:
@all-contributers please add @HugoMario for code and test

cdimascio:
@all-contributors please add @HugoMario for code and test

Issue #136, opened 2019-11-22
Query parameters including `[]` are not parsed correctly.


ckeboss:
If you have a parameter such as:
```
    FilterParam:
      name: filter[foo]
      in: query
      description: Filter foo
      required: false
      style: form
      explode: false
      schema:
        type: array
        items:
          type: string
```
which `?filter[foo]=1` would be valid, the validator responds with `Unknown query parameter filter` because it uses whichever query parser express is configured with. In this example, `req.query` would be:
```
{
  filter: {
    foo: 1
  }
}
```
which fails validation. I would propose to not use the express parsed query param and instead parse it independently to handle this.

cdimascio:
Will dig into this some. Thanks for the issue

ckeboss:
I have a proposed solution. Will make a PR later today

PR #142, opened 2019-11-25
Support a wider range of query param names

ckeboss:
Fixes https://github.com/cdimascio/express-openapi-validator/issues/136


ckeboss:
@cdimascio looks like the travis build has an issue?

```
The command "openssl aes-256-cbc -K $encrypted_deba340dfe89_key -iv $encrypted_deba340dfe89_iv -in secrets.zip.enc -out secrets.zip -d" failed and exited with 1 during .
```

src/middlewares/openapi.request.validator.ts
```diff
@@ -129,9 +129,27 @@ export class RequestValidator {
 
     const validator = this.ajv.compile(schema);
     return (req: OpenApiRequest, res: Response, next: NextFunction): void => {
+      const queryIndex = req.originalUrl.indexOf('?');
```
cdimascio: perhaps, move all of this query parse logic into a function e.g. `const query = parserQuery(..)`

cdimascio:
@ckeboss thanks for the PR!

cdimascio:
@ckeboss a made a minor [modification](https://github.com/cdimascio/express-openapi-validator/pull/142/commits/3a66fd7cc2bc31a901443580fb6690f1ea34a1da) that resolves the travis issue

ckeboss:
Made the updates requested. Do you think there are any unintended consequences of this change? I am wondering if anyone is relying on the express parsing behavior. I guess time will tell.

cdimascio:
@ckeboss this is a good question. i need to think on this a bit. most user's will certainly expect the expect parser given that is what they are working with when building apps. the express parser is also well-trodden and known to be solid. i think at best this needs be behind a feature flag.

question: when using express how do you handle your query syntax? are you using `req.query` or are you parsing the query params manually

ckeboss:
@cdimascio on second thought, this should not affect the parsing, as it will also parse `req.query` the same. It only validates the query params we parse though.

In express I use the default (https://www.npmjs.com/package/qs) query param parser.

Let me know if you have any ideas on more tests that could be written to verify this.

ckeboss:
Thoughts on this? As I said, I don't think this will affect anyone, as we parse both

cdimascio:
will revisit this soon

src/middlewares/openapi.request.validator.ts
```diff
@@ -341,4 +363,25 @@ export class RequestValidator {
...
+    const queryString = (queryIndex >= 0) ? url.slice(queryIndex + 1) : '';
+    const searchParams = new URLSearchParams(queryString);
+    const queryParamsToValidate: object = {};
+
+    searchParams.forEach((value, key) => {
+      if (queryParamsToValidate[key]) {
+        if (queryParamsToValidate[key] instanceof Array) {
+          queryParamsToValidate[key].push(value)
+        } else {
+          queryParamsToValidate[key] = [queryParamsToValidate[key], value]
```
cdimascio: @ckeboss, can you describe this case?

src/middlewares/openapi.request.validator.ts
```diff
@@ -153,6 +157,11 @@ export class RequestValidator {
        * https://swagger.io/docs/specification/describing-parameters/#schema-vs-content
        */
       parameters.parseJson.forEach(item => {
+        if (item.reqField === 'query' && queryParamsToValidate[item.name]) {
```
cdimascio: @ckeboss seems this can cause two calls to `JSON.parse(..` when handling `query` parameters. both on line `161` and `166`

src/middlewares/openapi.request.validator.ts
ckeboss: `URLSearchParams` does not do any sort of processing like some other search param parsers do.

So given the query string `test=foo&test=bar`, it would be parsed as

```
key: test, value: foo
key: test, value: bar
```

We check to see if the key has already been seen, and if it has, we parse all the same keys as an array. If we did not do this, the value of the key would continue to be overwritten, and only the last iteration would be saved.

src/middlewares/openapi.request.validator.ts
ckeboss: This is done for backwards compatibility. We continue to parse as we did before, mutating the `req` object.

But for actual query validation, we mutate the `queryParamsToValidate` that then is passes to `Ajv` for validation.

cdimascio: i'd like to avoid the second JSON.parse if possible. for example, if we have a value for `queryParamsToValidate[item.name]` that's already been `JSON.parse`d, can we assign that to `req[item.reqField][item.name]` instead of performing the `JSON.parse` again

cdimascio:
> Thoughts on this? As I said, I don't think this will affect anyone, as we parse both

agreed. let's do it

src/middlewares/openapi.request.validator.ts
ckeboss: Makes sense. I have added an equality check so if they two strings are equal, it will assign them the same parsed value and return. I do this with the string types. I did not for the array, as the extra step in preforming a deep equality check I don't think is worth it?

cdimascio:
@ckeboss I reverted this commit from master

A test that @JacobLey as part of this [PR](https://github.com/cdimascio/express-openapi-validator/pull/187) exposed a problem we'll need to solve for before this is merged in

unfortunately, because it was merged and reverted, github requires that a new PR be created. would you mind resubmitting a this PR and we can look into how to resolve this last remaining issue

apologies. the change is close. glad Jacob created this test which caught it

ckeboss:
@cdimascio yes, I will take a look.

ckeboss:
After looking at this, I think the best course of action is to document that if you want to use the validator with query params that will be parsed by `qs`, you need to set express to use either `simple` or `false`.

I'll make a PR for the documentation.

cdimascio:
@ckeboss Thanks so much for your investigation. This sounds good

nampas:
@ckeboss @cdimascio I'm trying to follow along here but I'm a tad confused.

I have a query param array that is supposed to be passed as separate duplicate keys, i.e. `?foo[]=1&foo[]=2&foo[]=3`. the parameter is specified as such:
```
    "foo": {
      "in": "query",
      "name": "foo[]",
      "schema": {
        "type": "string"
      },
      "explode": true,
   }
```
The OpenAPI validation is failing:
```
request.query['foo[]'] should be string
...
"variables":{"value":"[\"292017\",\"7754559\"]"
```
Any idea how to fix this?

ckeboss:
@nampas, did you change the express query parser to not use qs (extended)?

nampas:
@ckeboss Thanks for the response.

Sorta. I do need the `qs` parsed query elsewhere in my app, so I basically wrapped the `express-openapi-validator` middleware in my own custom middleware, which replaces the value of `request.query` with something that `express-openapi-validator` likes (i.e. query params parsed simple). Then my custom middleware restores the standard express `qs` parsed params after its finished executing. Pretty janky.

For the question I asked above, I think I may have figured it out. I needed to specify that the query params were arrays in the openapi spec 🤦 . The spec now looks like:
```
    "foo": {
      "in": "query",
      "name": "foo[]",
      "schema": {
        "type": "array",
        "items": {
          "type": "string"
        }
      },
      "explode": true,
   }
```
It would be nice to revisit your solution here. Ideally, I think, this library wouldn't rely on the query to be parsed any particular way. Grabbing the raw query from `originalUrl` and processing it seem to make a lost of sense.

Issue #143, opened 2019-11-25
Add support for content all content types


ckeboss:
I noticed the `TODO` comment for `add support for content all content types`.

Is there a current idea on how to do this? I would be happy to do the work.

cdimascio:
awesome! i very much welcome the help. thanks

all in all, this area hasn't been deeply explored. some types may just work, others may require a bunch of work.

as a starting point, have a look https://swagger.io/specification/#mediaTypeObject, particularly the Media Type Object and Media Type Examples sections

as for ways to get started:
- create some tests using different mime types and observe how the validator responds
- review some places in the code that lightly test media types other than json e.g. text/plain [here](https://github.com/cdimascio/express-openapi-validator/blob/bbcd5aed1f8e3fd872b1c88ac6756c572c4d7750/test/request.bodies.ref.spec.ts#L44). 
- multipart/formdata has some [tests](https://github.com/cdimascio/express-openapi-validator/blob/bbcd5aed1f8e3fd872b1c88ac6756c572c4d7750/test/multipart.spec.ts) due to file upload support
- binary media types - file upload support may need some enhancements as we currently hardcode `multipart/formdata` [here](https://github.com/cdimascio/express-openapi-validator/blob/bbcd5aed1f8e3fd872b1c88ac6756c572c4d7750/src/middlewares/openapi.multipart.ts#L54). For example, some modifications may be needed to support the following:
```
requestBody:
  content:
      # a binary file of type png or jpeg
    'image/jpeg':
      schema:
        type: string
        format: binary
    'image/png':
      schema:
        type: string
        format: binary     
```
-  application/x-www-form-urlencoded. does this just work? does it need work? note we hard code `application/json` in a few places ([1](https://github.com/cdimascio/express-openapi-validator/blob/bbcd5aed1f8e3fd872b1c88ac6756c572c4d7750/src/middlewares/ajv/index.ts#L103), [2](https://github.com/cdimascio/express-openapi-validator/blob/bbcd5aed1f8e3fd872b1c88ac6756c572c4d7750/src/middlewares/openapi.request.validator.ts#L331), [3](https://github.com/cdimascio/express-openapi-validator/blob/bbcd5aed1f8e3fd872b1c88ac6756c572c4d7750/src/middlewares/openapi.response.validator.ts#L105)) which may need to change
```
    application/x-www-form-urlencoded:
      schema:
        type: object
        properties:
          id:
            type: string
            format: uuid
          address:
[...]

PR #147, opened 2019-11-27
WIP: Allow different json media types for params, requests, and responses

ckeboss:
Addresses https://github.com/cdimascio/express-openapi-validator/issues/143


src/middlewares/ajv/index.ts
```diff
@@ -95,17 +93,5 @@ function createAjv(
     );
   }
 
-  if (openApiSpec.components.requestBodies) {
```
cdimascio: We shouldn't remove this addSchema section as we need to add shemas any `components.requestBodies` 
for references: https://swagger.io/docs/specification/components/

ckeboss: I believe this is handled by AJV when we pass the request body.

Also, all the test work without it? Is there a specific test I could add to verify this?

ckeboss:
We could use both https://www.npmjs.com/package/media-typer and https://www.npmjs.com/package/content-type together, as they do the same thing, but split into multiple libraries.

cdimascio:
Cool. Let's go with  https://www.npmjs.com/package/media-typer and https://www.npmjs.com/package/content-type

src/middlewares/ajv/index.ts
cdimascio: Good point. You're right. This is tested in request.bodies.ref.spec.ts. Looks like its safe to remove

cdimascio:
@ckeboss i've merged in the v3.0.0 codebase into your PR branch.

It will be fantastic if you can make the changes to use [media-types](https://www.npmjs.com/package/media-typer) and [content-type](https://www.npmjs.com/package/content-type) instead of media-type.

Looking forward to getting this in!
Thanks!

ckeboss:
Thanks. Yes, I'll make those changes today

cdimascio:
@ckeboss i've merged master into your PR

ckeboss:
Thank you, sorry, been busy, will change out those libraries soon.

ckeboss:
@cdimascio updated the media type / content type parser libraries.

cdimascio:
Thanks a lot @ckeboss. Will have a look soon!

cdimascio:
@all-contributors add @ckeboss for code and test

Issue #159, opened 2019-11-28
Could not locate bindings file


stanleynguyen:
I've just upgraded to `2.18.0` and faced with this issue

```
    Could not locate the bindings file. Tried:
     → /Users/stanleynguyen/Documents/Projects/template-node/{{cookiecutter.service_name}}/node_modules/deasync/build/deasync.node
     → /Users/stanleynguyen/Documents/Projects/template-node/{{cookiecutter.service_name}}/node_modules/deasync/build/Debug/deasync.node
     → /Users/stanleynguyen/Documents/Projects/template-node/{{cookiecutter.service_name}}/node_modules/deasync/build/Release/deasync.node
     → /Users/stanleynguyen/Documents/Projects/template-node/{{cookiecutter.service_name}}/node_modules/deasync/out/Debug/deasync.node
     → /Users/stanleynguyen/Documents/Projects/template-node/{{cookiecutter.service_name}}/node_modules/deasync/Debug/deasync.node
     → /Users/stanleynguyen/Documents/Projects/template-node/{{cookiecutter.service_name}}/node_modules/deasync/out/Release/deasync.node
     → /Users/stanleynguyen/Documents/Projects/template-node/{{cookiecutter.service_name}}/node_modules/deasync/Release/deasync.node
     → /Users/stanleynguyen/Documents/Projects/template-node/{{cookiecutter.service_name}}/node_modules/deasync/build/default/deasync.node
     → /Users/stanleynguyen/Documents/Projects/template-node/{{cookiecutter.service_name}}/node_modules/deasync/compiled/10.16.0/darwin/x64/deasync.node
     → /Users/stanleynguyen/Documents/Projects/template-node/{{cookiecutter.service_name}}/node_modules/deasync/addon-build/release/install-root/deasync.node
     → /Users/stanleynguyen/Documents/Projects/template-node/{{cookiecutter.service_name}}/node_modules/deasync/addon-build/debug/install-root/deasync.node
     → /Users/stanleynguyen/Documents/Projects/template-node/{{cookiecutter.service_name}}/node_modules/deasync/addon-build/default/install-root/deasync.node
     → /Users/stanleynguyen/Documents/Projects/template-node/{{cookiecutter.service_name}}/node_modules/deasync/lib/binding/node-v64-darwin-x64/deasync.node

      at bindings (node_modules/bindings/bindings.js:126:9)
      at Object.<anonymous> (node_modules/deasync/index.js:30:31)
[...]

Issue #151, opened 2019-11-27
Endless loop caused by json-schema-ref-parser


scriptify:
Hello!

For some strange reason I yet not know, the "bundle" method of the class "$RefParser" seems to cause an endless loop on line 16: https://github.com/cdimascio/express-openapi-validator/blob/b8e34304ab5c2821a51c986e43b4b6ae4db51d05/src/framework/json.ref.schema.ts#L16

As I am the first one with this issue, maybe it is caused by my setup.

Those are my dependencies:

```json
{
    "@typegoose/typegoose": "^6.1.5",
    "cors": "^2.8.5",
    "express": "^4.17.1",
    "express-openapi-validator": "^2.17.2",
    "express-session": "^1.17.0",
    "mongoose": "^5.7.12",
    "openid-client": "^3.8.3",
    "react": "^16.11.0",
    "react-dom": "^16.11.0",
    "swagger-ui-express": "^4.1.2",
    "webpack-pwa-manifest": "^4.0.0",
    "workbox-webpack-plugin": "^3.6.3",
    "yamljs": "^0.3.0"
  }
```

I also tried using different OpenAPI specs (also the one in the tests directory of this repo), with the same result.

Maybe it's a dependecy conflict somehow?

scriptify:
Update: I don't know why yet, but it has to do something with the database connection I'm establishing, which seems very strange. If I establish the MongoDB connection _after_ calling **install**, it works just fine. Otherwise, it keeps blocking the whole thread. I am not sure why this is happening.

cdimascio:
@scriptify this change is part of >2.17.0 to enable support spec splitting via $refs. the change has been live for 2days, so other may run into this.

based on the code snippet you provided, this is coming from the use of package `deasync`. i used this package (despite the risks) to synchronously resolve all `$refs` in a multiple file spec. i did this in order to preserve the current api. using `deasync` to force synchronicity appears to be the root of this problem. i'm not sure if there is a good solution to fix this as is. it may require handling the initial parse asynchronously, which ultimately means the api for express-openapi-validator must change and a new major version must be rolled.

in any case, i'll start investigating that path.
if you find a workaround, please let me know. i'll certainly document it or provide a patch. 
if you find a solution, PRs are always welcome!

cdimascio:
Also, if you can provide a simple express app that demonstrates the problem, I'll dig in.

cdimascio:
[...]

PR #161, opened 2019-11-29
v3.0.0-alpha.1 - parse api spec asynchronously 

cdimascio:
@richdouglasevans @gonenduk @sheldhur @scriptify @comino would very much appreciate your feedback on the following. I am considering changing the API in v3 and making the lib fully async.

To see how usage might change see
- The Usage section of the [README](https://github.com/cdimascio/express-openapi-validator/blob/fc27c2cf92daf1b7d90ff0974374dd66affe3cd1/example/README.md)
- The [example app](https://github.com/cdimascio/express-openapi-validator/blob/fc27c2cf92daf1b7d90ff0974374dd66affe3cd1/example/app.js#L26)

The motivation for this PR is to remove the dependency on the [deasync](https://github.com/abbr/deasync) library. In order to do so, the parsing and json-ref resolution (spec splitting) will now occur asynchronously. This impacts the API surface as the `install` method can no longer be syncrhonous. Ultimately, this means routes/middleware functions must be applied once the promise settles (async/await can be used) or via a callback.

**Notes:**
- Remove dependency on `deasync`
- These changes modify the API surface and constitute a breaking change

  **`install`** returns a Promise<Express.Application> instead of void

  **old**
  ```javascript
  new OpenApiValidator({
    apiSpec: './example.yaml',
  }).install(app);

  // custom routes
  app.get(...)

  app.use(errorHandler)
  // ...
  ```

  **new** 
  ```javascript
  new OpenApiValidator({
    apiSpec: './example.yaml',
  })
  .install(app)
  .then(() => {
    // custom routes
    app.get(...)

    app.use(errorHandler)
    // ...
  });
  ```


scriptify:
I think removing deasync will be a huge win! There's just one thing I not sure about: Using it in a serverless way. If someone for example uses [aws-serverless-express](https://github.com/awslabs/aws-serverless-express) or something similar, there could be a problem with routes not being setup when the request comes in. Any thoughts about this? Is this library even inteded to be used in a serverless way? I could do a quick test if so.

cdimascio:
this will not be a problem. node behaves in an asynchronous manner, hence all js based libs, frameworks, etc do as well. all in all, there is nothing technical that prevents this from working in a serverless environment.

note that endpoints to be registered after the install's promise resolves e.g. in `.then` (see below). because of this, it is impossible for one these endpoints to handle a request prior to validator setup/install. in the absolutely worst case, a request(s) that comes in prior to validation being initialized will be queued on the promise. 

```javascript
new OpenApiValidator({
  apiSpec: './example.yaml',
})
.install(app)
.then(app => {
  // add routes to validate here <----------
});
```

hope this helps

cdimascio:
on a tangential note, i can support both sync and async behavior

**Async**
To install:
```shell
npm install express-openapi-validator
```

To use:

```javascript
await new OpenApiValidator({
  apiSpec: './example.yaml',
}.install(app)
```

**Sync**
To install:
```shell
npm install express-openapi-validator deasync
```

To use:

```javascript
new OpenApiValidator({
  apiSpec: './example.yaml',
}.installSync(app)
```

Thoughts?

scriptify:
Yea that's what I meant, I'm not doubting that it works in a serverless environment, but requests may come in prior the installation, because serverless functions (at least when cold starting) don't wait for your endpoints being available.

I like the option of 2 different install modi! People could opt-in for the sync version if the asynchronicity is causing any problems.

cdimascio:
@scriptify 

a version of v3 alpha is available on npm. please give it a try. it makes `deasync` optional in that it must be installed explicitly. it is only needed if using the lib synchronously.

The alpha version is available now: "express-openapi-validator": "^3.0.0-alpha.1"

See the [alpha readme](https://github.com/cdimascio/express-openapi-validator/tree/async_1)) for details

cdimascio:
Addresses #162, #159, and #151

gonenduk:
Since deasync is a peer - not installing it gives a warning in npm:
`npm WARN express-openapi-validator@3.0.0-alpha.2 requires a peer of deasync@^0.1.16 but none is installed. You must install peer dependencies yourself.`
Is there a way to make it optional so we don't get the error?

cdimascio:
Unfortunately, there is not a good solution. Npm, `optionalDependencies` are automatically installed unless the user explicitly **opts-out** by specifying `npm install --no-optional`. `peerDependencies` (which I'm currently using)  are **opt-in** which is nice, but emit the warning (not so nice). Ultimately, we want `optionalPeerDependencies`, but that does not exist.

cdimascio:
looks like a peer dep that is not installed can cause problems with shrinkwrap and potentially other tools. i'm going to drop support for synchronous install. if there is demand i'll consider it as a supplementary package.

PR #169, opened 2019-12-04
Opt-in multer

yskeat:


yskeat:
I will take a look at the test

yskeat:
After read multer source code, I will say that based on how multer implement, i think my approach is not possible. 
I will call off for this pull request.

cdimascio:
What did your discover that makes this not possible?

yskeat:
As I understand and tried, multer is needed to process multipart/form-data which is what the original code did it correctly.
If we skip the multer.any() call, req.body will always be an empty object if request type is multipart.
The way multer implement is the code to extract field and file from the form and put it into req.body and req.file is in the same function body.

PR #184, opened 2019-12-16
Fix $ref issue  with special characters (#183)

### LINKED ISSUE #183: Issue with some $ref
Hello,

I saw it is possible to split specifications into multiple files. 
I tried to split mine ( [from my master branch](https://raw.githubusercontent.com/SourceCodeOER/sourcecode_api/master/api.yml) ) into a [separate branch](https://github.com/SourceCodeOER/sourcecode_api/tree/delete_exercises_and_tags) : the ["main file"](https://raw.githubusercontent.com/SourceCodeOER/sourcecode_api/delete_exercises_and_tags/api.yml) and the [rest](https://github.com/SourceCodeOER/sourcecode_api/tree/delete_exercises_and_tags/openapi)

( I didn't forget to escape special characters like explained in Swagger : https://swagger.io/docs/specification/using-ref/ or in this [old open api v2 tutorial](https://apihandyman.io/writing-openapi-swagger-specification-tutorial-part-8-splitting-specification-file/#modifying-paths-references) )

To test if the specifications are correct (and work as expected), I used the following commands explained in my readme : 

image

The validator seems to accept the specifications : 

image

However, some endpoints like /auth/login and /auth/register no longer work and send me an error of this type : 

image

Any idea what caused this problem? The path seems to be recognized in other tools ( like the OpenAPI (Swagger) Editor extension in Visual Studio Code ) 

image

Thanks

jy95:
Attempt to fix #183 .
In my opinion ( as some of the tests uses circular structure ) , it is better to give choice to final user to use dereference / bundle if they got a similar situation like mine.

For example : 
```js
new OpenApiValidator({
    unsafeRefs: true
}).install(app)
```

### DIFF
--- a/src/framework/index.ts
+++ b/src/framework/index.ts
@@ -71,7 +71,7 @@ export class OpenAPIFramework {
     };
   }
 
-  private loadSpec(filePath: string | object, $refParser: { mode: 'bundle' | 'dereference' }): Promise<OpenAPIV3.Document> {
+  private loadSpec(filePath: string | object, $refParser: { mode: 'bundle' | 'dereference' } = { mode: 'bundle'}): Promise<OpenAPIV3.Document> {
     // Because of this issue ( https://github.com/APIDevTools/json-schema-ref-parser/issues/101#issuecomment-421755168 )
     // We need this workaround ( use '$RefParser.dereference' instead of '$RefParser.bundle' ) if asked by user
     if (typeof filePath === 'string') {
--- a/src/framework/index.ts
+++ b/src/framework/index.ts
@@ -71,7 +71,9 @@ export class OpenAPIFramework {
     };
   }
 
-  private loadSpec(filePath: string | object, $refParser: { mode: 'bundle' | 'dereference' } = { mode: 'bundle'}): Promise<OpenAPIV3.Document> {
+  private loadSpec(
+    filePath: string | object,
+    $refParser: { mode: 'bundle' | 'dereference' } = { mode: 'bundle' }): Promise<OpenAPIV3.Document> {
     // Because of this issue ( https://github.com/APIDevTools/json-schema-ref-parser/issues/101#issuecomment-421755168 )
     // We need this workaround ( use '$RefParser.dereference' instead of '$RefParser.bundle' ) if asked by user
     if (typeof filePath === 'string') {
@@ -86,7 +88,7 @@ export class OpenAPIFramework {
             fs.readFileSync(absolutePath, 'utf8'),
             { json: true },
           );
-          return ($refParser.mode === "bundle") ? $RefParser.bundle(docWithRefs) : $RefParser.dereference(docWithRefs);
+          return ($refParser.mode === 'bundle') ? $RefParser.bundle(docWithRefs) : $RefParser.dereference(docWithRefs);
         } finally {
           process.chdir(origCwd);
         }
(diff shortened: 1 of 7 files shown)


src/framework/index.ts
```diff
@@ -23,7 +23,7 @@ export class OpenAPIFramework {
     visitor: OpenAPIFrameworkVisitor,
   ): Promise<OpenAPIFrameworkInit> {
     const args = this.args;
-    const apiDoc = await this.copy(await this.loadSpec(args.apiDoc));
+    const apiDoc = await this.copy(await this.loadSpec(args.apiDoc, args.unsafeRefs));
```
cdimascio: would it be possible to use dereference instead of bundle, yet use [lodash.clonedeep](https://www.npmjs.com/package/lodash.clonedeep) to copy the spec? it would be great if we can remove the need for an option

jy95: Give me your full code proposal and I will test on my project to tell you if it works.
If not, I am afraid it is the only solution : bundle works great on detecting circular refs ( like some of your tests ) but doesn't work with escaped paths ...

cdimascio: the doc in dereference mentions
>  JSON Schema files can contain circular $ref pointers, and JSON Schema $Ref Parser fully supports them. Circular references can be resolved and dereferenced just like any other reference. However, if you intend to serialize the dereferenced schema as JSON, then you should be aware that JSON.stringify does not support circular references by default, so you will need to use a custom replacer function.
>
>You can disable circular references by setting the dereference.circular option to false. Then, if a circular reference is found, a ReferenceError will be thrown.
>
>Or you can choose to just ignore circular references altogether by setting the dereference.circular option to "ignore". In this case, all non-circular references will still be dereferenced as normal, but any circular references will remain in the schema.
>
>Another option is to use the bundle method rather than the dereference method. Bundling does not result in circular references, because it simply converts external $ref pointers to internal ones.

We end up serializing the `apiDoc` with JSON.stringify(). The `copy` method on line 26 does this. I wonder will it work with `dereference` if we change the way `copy` works. For example, we could use `lodash.deepCopy` or [JSON.stringify with a replacer function](https://stackoverflow.com/questions/11616630/how-can-i-print-a-circular-structure-in-a-json-like-format). If we can use dereference always we can avoid the need for an option. It seems to make dereference work we need to avoid serializing `apiDoc` since it may have circular refs

jy95: It is worth a try : if this passes my example of specification above then we can consider the problem solved.

jy95:
@cdimascio I add a test to reproduce the bug - You can try what you think in order to solve the bug

src/framework/index.ts
cdimascio: 🤞 let me know how it turns out

jy95: You can fork this branch and try : I am not completly sure how you would like to change the code to handle this situation

cdimascio: deepclone doesn't help, unfortunately. appears we indeed require an option

perhaps instead of `$unsafeRefs`(since dereference doesn't necessarily mean refs are unsafe), we provide an option to use `bundle` or `dereference`. its a bit of an implementation detail that ends up getting exposed, but perhaps we can remove this option at some point in the future (if json-ref-parser provides an alternative)

Perhaps something like:
```javascript
  $refParser: {
     mode: 'bundle' | 'dereference
  }
```
where  $refParser.mode = 'bundle' is the default when not present

thoughts?

jy95: LGTM : Feel free to rename as you wish

jy95:
Sorry for the many commits : my linter rules aren't the same that Codacy ( you can squash them before merging )

src/framework/index.ts
cdimascio: will look at this shortly and get this in

Issue #186, opened 2019-12-17
anyOf/oneOf/allOf should not imply type=object


JacobLey:
Hello,

I think I've hit an unintentional backwards incompatibility caused by this PR/commit: https://github.com/cdimascio/express-openapi-validator/pull/126/commits/bc44495375eb730becd602d3b1f0bf75fb2d3c1e

In my case, I have an input that I accept as a number or a string (using `allOf`) to represent a timestamp.  When provided a string, the JSON parsing fails, but I believe it should not be parsed in the first place, as I did not indicate it was an object.

Parsing is performed here: https://github.com/cdimascio/express-openapi-validator/blob/master/src/middlewares/openapi.request.validator.ts#L156

This example route should help clarify:

```
/my/get/endpoint:
  get:
    operationId: my.get
    summary: Failing Timestamp Get
    description: This endpoint will fail if query's "time" is an ISO string
    parameters:
      - name: time
        in: query
        description: Limit results to ones created before time
        schema:
          title: Input Timestamp
          description: Value passed to Javascript's `new Date()`.
          example: '2019-06-24T12:34:56.789Z'
          anyOf:
            - title: Number Timestamp
              type: integer
              description: Unix milliseconds
              example: 1234567890123
              nullable: true
            - type: string
              description: ISO Timestamp
              pattern: \S
    responses:
      200:
        description: My Response
        content:
        application/json:
          schema:
            type: object
```

When time is `2019-06-24T12:34:56.789Z` I get an error `Unexpected number in JSON at position 4`

Based on the OpenAPI spec I don't think I would expect a JSON.parse to be attempted here. Unless I am incorrect? In which case I'd appreciate some guidance around how to properly format my schema.

I believe we should either only be parsing when type = object, or one of the any/one/all is of type object.

Or we could always attempt JSON.parse, but not fail on error (i.e. wrap in try/catch) and let the validator detect if parsing should have succeeded (e.g. non-json string doesn't match `object` schema)

The latter option feels better to me, as it would support both non-json+json strings for the same parameter.

I'm happy to take a swing on a PR for this if my suggested solution is agreeable.

cdimascio:
@JacobLey, thanks for jumping on this. im on board with your approach. Will review the PR soon.

cdimascio:
Thanks, @JacobLey. This is fixed in `v3.2.3`

PR #187, opened 2019-12-17
Only enforce query object parsing when specified, always validate #186

JacobLey:
This updated test+parsing highlights when forced JSON.parsing is not appropriate, and should rely on validator to determine success.

Opened as an attempt to solve this issue: https://github.com/cdimascio/express-openapi-validator/issues/186

I tried my best to match existing code style, but wasn't sure if there was an official linter command/test. If so that has not been run but I am happy to tweak if necessary.

### DIFF
--- a/src/middlewares/openapi.request.validator.ts
+++ b/src/middlewares/openapi.request.validator.ts
@@ -372,7 +372,7 @@ export class RequestValidator {
             [].concat(
               schema.allOf,
               schema.oneOf,
-              schema.anyOf,
+              schema.anyOf
             ).some(schemaHasObject)
           );
 
--- a/src/middlewares/openapi.request.validator.ts
+++ b/src/middlewares/openapi.request.validator.ts
@@ -366,21 +366,15 @@ export class RequestValidator {
         }
       } else if ($in === 'query') {
         // handle complex json types in schema
-        const schemaHasObject = schema => {
-          return (
-            schema.type === 'object' &&
-            schema.type !== 'array'
-          ) || (
-            schema.allOf &&
-            schema.allOf.some(schemaHasObject)
-          ) || (
-            schema.oneOf &&
-            schema.oneOf.some(schemaHasObject)
-          ) || (
-            schema.anyOf &&
-            schema.anyOf.some(schemaHasObject)
-          )
-        };
+        const schemaHasObject = schema =>
+          schema && (
+            schema.type === 'object' ||
+            [].concat(
+              schema.allOf,
+              schema.oneOf,
+              schema.anyOf,
+            ).some(schemaHasObject)
+          );
 
         if (schemaHasObject(parameterSchema)) {
           parseJson.push({ name, reqField });
--- a/src/middlewares/openapi.request.validator.ts
+++ b/src/middlewares/openapi.request.validator.ts
@@ -153,9 +153,14 @@ export class RequestValidator {
        */
       parameters.parseJson.forEach(item => {
         if (req[item.reqField]?.[item.name]) {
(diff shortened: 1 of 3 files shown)


JacobLey:
Ah looks like Travis does the linting

JacobLey:
I'm stumped on the linting... I can't figure out the run-local command so I don't pollute this PR with attempts.

In the last commit I was trying to replicate this test's usage of response.body: https://github.com/cdimascio/express-openapi-validator/blob/master/test/response.validation.spec.ts#L88

JacobLey:
Apologies for the linter noise, I can re-open a PR/force squash those commits to clean it up if necessary

src/middlewares/openapi.request.validator.ts
```diff
@@ -359,16 +364,27 @@ export class RequestValidator {
...
-          parameterSchema.oneOf ||
-          parameterSchema.anyOf ||
-          (parameterSchema.type === 'object' &&
-            parameterSchema.type !== 'array'))
-      ) {
-        parseJson.push({ name, reqField });
+        const schemaHasObject = schema => {
+          return (
+            schema.type === 'object' &&
+            schema.type !== 'array'
```
JacobLey: This was copy+paste... but realistically is unnecessary (tautology)

test/resources/serialized.objects.yaml
```diff
@@ -32,21 +33,41 @@ paths:
...
-                
-components:
-  schemas:
-    someItem:
-      type: object
-      properties:
-        id:
-          type: string
\ No newline at end of file
+                type: object
```
cdimascio: indentation looks off on this

test/resources/serialized.objects.yaml
```diff
@@ -32,21 +33,41 @@ paths:
...
+
+components: {}
```
cdimascio: is this necessary?

cdimascio:
> This updated test+parsing highlights when forced JSON.parsing is not appropriate, and should rely on validator to determine success.
> 
> Opened as an attempt to solve this issue: #186
> 
> I tried my best to match existing code style, but wasn't sure if there was an official linter command/test. If so that has not been run but I am happy to tweak if necessary.

in terms of linting. i use prettier to autoformat the code. `.prettierrc.json` is in the project root

test/resources/serialized.objects.yaml
JacobLey: Are you viewing diff as "unified"? Looks off there but with "split" I think it is correctly indented.

test/resources/serialized.objects.yaml
JacobLey: Removing it fails on this line:
https://github.com/cdimascio/express-openapi-validator/blob/master/src/framework/ajv/index.ts#L88

Not technically required by OpenAPI spec: https://github.com/OAI/OpenAPI-Specification/blob/master/schemas/v3.0/schema.json#L44

but this PR didn't seem like the appropriate place to address.

JacobLey: Adding `?` in the `ts` file would probably fix (might be a couple spots with similar issues), I could include if you want or open follow-on PR

cdimascio: ah yep, you're right. if you don't mind adding it in a follow on PR that would be fantastic

test/resources/serialized.objects.yaml
cdimascio: you're right i misread. looks good

cdimascio:
@all-contributors add @JacobLey  for code and test

PR #193, opened 2019-12-20
Proposal : fix object serialization when explode and form are used

### LINKED ISSUE #192: Postman - nested parameters for query
Hi,

I have a little concern with postman when it comes to deal with nested parameters ([swagger - parameter Serialization](https://swagger.io/docs/specification/serialization/))

There's no `settings` object generated (see screenshot below).

Capture d’écran 2019-12-19 à 20 15 04

It seems that this issue is related to #125.

When we set `explode: true` and `style: form`, it does the same as we can see on postman.

Is there a workaround on that ?

Thanks :)

jy95:
Attempt to fix #192 

In my opinion, it is the simplest way to correct this issue : as many openapi specifications uses "explode": true and "style": form only at the top level , no need to adapt the schema to deal with corner cases

### DIFF
--- a/src/middlewares/openapi.request.validator.ts
+++ b/src/middlewares/openapi.request.validator.ts
@@ -128,6 +128,25 @@ export class RequestValidator {
 
     const validator = this.ajv.compile(schema);
     return (req: OpenApiRequest, res: Response, next: NextFunction): void => {
+
+      // forcing convert to object if scheme describes param as object + explode
+      // for easy validation, keep the schema but update whereabouts of its sub components
+      parameters.parseObjectExplode.forEach(item => {
+        if (req[item.reqField]) {
+          // check if there is at least one of the nested properties before create the parent
+          const atLeastOne = item.properties.some(p => req[item.reqField].hasOwnProperty(p));
+          if (atLeastOne) {
+            req[item.reqField][item.name] = {};
+            item.properties.forEach(property => {
+              if (req[item.reqField][property]) {
+                req[item.reqField][item.name][property] = req[item.reqField][property];
+                delete req[item.reqField][property];
+              }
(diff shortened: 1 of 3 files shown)


src/middlewares/openapi.request.validator.ts
```diff
@@ -128,6 +128,25 @@ export class RequestValidator {
 
     const validator = this.ajv.compile(schema);
     return (req: OpenApiRequest, res: Response, next: NextFunction): void => {
+
+      // forcing convert to object if scheme describes param as object + explode
+      // for easy validation, keep the schema but update whereabouts of its sub components
+      parameters.parseObjectExplode.forEach(item => {
+        if (req[item.reqField]) {
```
jy95: Check may not be needed since express always has a property set to {} for each type of parameter. 
Prevention is better than cure ^^

cdimascio:
@jy95 thanks. this looks like a reasonable approach. note that there were a couple of use cases missing that i fixed. please review

also, the query parameter parsing logic is starting to get a bit unwieldy. it certainly needs a refactor

jy95:
@cdimascio LGTM. To inspire you, it seems this [validator](https://github.com/byu-oit/openapi-enforcer/blob/master/src/enforcers/Parameter.js#L133) has a more or less good way to deal with this mess.

cdimascio:
@jy95 i've refactored the parameter validator. it's now colocated to this [file](https://github.com/cdimascio/express-openapi-validator/blob/master/src/middlewares/parsers/req.parameter.mutator.ts) and is far more manageable

jy95:
@cdimascio Nice, it looks like way better. Maybe a day I will reuse your validator 👍

PR #211, opened 2020-01-09
Fix for nested routes validation

vdmitriy:
Validator doesn't work if it installed for nested route instead root app, because  when called from a middleware, the mount point is not included in `req.path`

### DIFF
--- a/src/middlewares/openapi.metadata.ts
+++ b/src/middlewares/openapi.metadata.ts
@@ -29,7 +29,7 @@ export function applyOpenApiMetadata(
   };
 
   function lookupRoute(req: OpenApiRequest): OpenApiRequestMetadata {
-    const path = req.path;
+    const path = req.originalUrl.split("?")[0];
     const method = req.method;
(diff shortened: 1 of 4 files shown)


cdimascio:
@all-contributors add @vdmitriy for code and test

cdimascio:
@vdmitriy your fix is available in `v3.6.1`. thanks again

cdimascio:
@vdmitriy given this new change, it appears the `install` methods should have their type signatures updated to indicate that either an express `Application` or an express `Router`. 

is this correct?

Assuming so, would you mind updating the signatures on the 3 `install` methods.

For example, in `src/index.ts` the type signatures should indicate the acceptable params
```typescript
  public async install(app: Application | Router): Promise<void>;
  public install(app: Application | Router, callback: (error: Error) => void): void;
  public install(app: Application | Router, ...
```

Note that making those changes will affect some of downstream call sites and functions and will require updating their types as well.

Finally, the readme should likely indicate that either app or router can be passed to the various install methods.

PR #221, opened 2020-01-16
Update deprecated multerOpts reference, update test

JacobLey:
Adding failing test first. Will follow up with fix + link to issue

### DIFF
--- a/test/multipart.spec.ts
+++ b/test/multipart.spec.ts
@@ -7,23 +7,38 @@ import * as packageJson from '../package.json';
 
 describe(packageJson.name, () => {
   let app = null;
+  const fileNames = [];
   before(async () => {
     const apiSpec = path.join('test', 'resources', 'multipart.yaml');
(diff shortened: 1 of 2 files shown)


test/multipart.spec.ts
```diff
@@ -69,8 +84,9 @@ describe(packageJson.name, () => {
             .to.have.property('fieldname')
             .to.equal('file');
           expect(b.metadata).to.equal('some-metadata');
-        }));
-
+        });
+      expect(fileNames).to.deep.equal(['package.json']);
```
JacobLey: Not sure what the best pattern here is (e.g. should fileNames really be unchanged on all tests, and cleared before each? Or maybe it somehow only exists in this test?)
The "filter" function just seemed like an easy way to check that multer is actually receiving the options + file

cdimascio: this is fine for now. i don't have better idea atm. we can improve the test later if needed

PR #228, opened 2020-02-09
wip: foway0's initial pass (see PR #229 instead)

cdimascio:


cdimascio:
@foway0 great work! this is fantastic!

i have a couple questions:

i see you created two new vendor extensions `x-eov-operationId` and `x-eov-controller`. 

could you describe the reasoning behind creating `x-eov-operationId`, rather than using `operationId`

I do like the fact that both fields are vendor extensions i.e. have the prefix `x-eov-`. It makes it very clear to the user that both are used by the validator.

On the other hand, I wonder if `x-eov-operationId` always had the same value as `operatorId` would everything work properly? if so, perhaps `x-eov-operationId` is redundant. thoughts?

cdimascio:
@nagarajanceg, i know that you are interested in this feature. i would love to get your thoughts as well. let me know if you have any thoughts with respect to my question above (or anything else) thanks!

cdimascio:
Perhaps it makes sense to have both `operationId` and `eov-operation-id`. Both would be optional, but at least one is required.

Given this, it would work as follows:

- `operationId` is used to identify the operation id. This enables folks who may already by using OpenAPIs `operationId` for other tools to also use it with eov. 

- if eov-operation-id` is specified it overrides `operationId`. This is beneficial from a code style perspective. For example, `operationId` must be globally unique, while `eov-operation-id` must only be unique per controller module. Thus, if we have a file `pets.js`, the `eov-operation-id` may can be unqualified e.g. `get`, `create`, `delete`. On the other hand, OpenAPI 3 requires `operationId` to be globally unique, thus methods must contain possibly redundant qualifiers (or names not idiomatic to javascript.

All in all, I'm feeling compelled to support both.
`operationId` to support existing OpenAPI workflows
`eov-operation-id` to enable users more control

cdimascio:
@foway0 , @nagarajanceg, i've made some minor modifications to @foway0's code. I've also put together some documentation.

Please see the new pull request here ->>>  https://github.com/cdimascio/express-openapi-validator/pull/229 for latest changes

@foway0 @nagarajanceg are you interested in trying it out? If so, I will publish an alpha version to npm. It will be great to get your thoughts prior to integrating into master

foway0:
First, thank you for the review.

When I was designing, I thought:

By oas document, Some tools use this value to define the name of the method.
Thus, I thought it could cause a crash when use other tools together.
That's why i prepared two new vendor extensions(`x-eov-operataion-id`, `x-eov-controller`).


As you said, `x-eov-operation-id` and `operationId` may have the same value.
So it seems better to choose either `x-eov-operation-id` AND `operationId` depending on the situation.
It will be better than my ideas! 


I thought there was a problem writing the README with a Google translator. :)
Thank you very much for your support. @cdimascio!


If you publish an alpha version, I will try it on my project! Thank you very much!

PR #251, opened 2020-03-02
Update openapi.request.validator.ts

mirekgw:
Ignore body validation when body content type is 'application/octet-stream'

### DIFF
--- a/src/middlewares/openapi.request.validator.ts
+++ b/src/middlewares/openapi.request.validator.ts
@@ -92,8 +92,10 @@ export class RequestValidator {
     const securityQueryParam = Security.queryParam(apiDoc, reqSchema);
     const body = bodySchemaParser.parse(path, reqSchema, contentType);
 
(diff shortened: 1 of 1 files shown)


cdimascio:
would something like the following resolve your use case?
```typescript
    const isBodyBinary = body?.['format'] === 'binary';
    const properties: ValidationSchema = { ...parameters, body: isBodyBinary ? {} : body };
    // TODO throw 400 if missing a required binary body
    const required = (<SchemaObject>body).required && !isBodyBinary ? ['body'] : [];
```
essentially, the same idea but ignore the body check if its formatted as binary. The presumption is that a simliar issue will occurs for other binary types e.g. `image/png`

cdimascio:
@mirekgw i tested this modification on your sample project. it all looks good. will get this merged in asap

cdimascio:
you fix is in `3.9.3` please give it a try

cdimascio:
@all-contributors add @mirekgw for code

PR #255, opened 2020-03-04
Update package.json

AlAyoub:
Revert to a secure version of multer (2.0.0-beta.1). Multer 1.4.2 has a security issue per the Sonatype Nexus IQ and I am proposing this change.


cdimascio:
@AlAyoub looks like this update broke some tests.

cdimascio:
Also, note that if you disable, `fileUploader: false` multer will not be used. This may provide you a workaround.

PR #258, opened 2020-03-07
Changed validator.default to validators.default in openapi.response.validator.ts

Emmeral:
The wrong variable was used to access the validator which led to an error. Now the default validator can be accessed.


cdimascio:
@Emmeral thanks for catching this and for providing the fix

cdimascio:
@all-contibutors add @Emmeral for code

cdimascio:
@all-contributors add @Emmeral for code

cdimascio:
fix is in `3.9.5`

PR #267, opened 2020-04-01
Features/allow full format validation

jakubskopal:
Allow to specify `format` option to AJP to be able to validate `date-time` format fully. By default only the syntax is checked. With `formatValidation='full' AJP also validates, that the date actually exists.
Default behaviour is compatible with past version.

### DIFF
--- a/README.md
+++ b/README.md
@@ -665,6 +665,14 @@ Determines whether the validator should coerce value types to match the type def
 - `false` - no type coercion.
 - `"array"` - in addition to coercions between scalar types, coerce scalar data to an array with one element and vice versa (as required by the schema).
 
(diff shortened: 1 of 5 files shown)


jakubskopal:
There is new option to `OpenApiValidator` called `formatValidation` which correspond to `format` property of AJP.

Allowable values are:
- false: do not check formats at all
- `fast`: only check syntax
- `full`: check also semantics (illegal date)

jakubskopal:
I've renamed it accordingly and added description to `README.md`

cdimascio:
@all-contributors add @jakubskopal for code, test, and documentation

Issue #271, opened 2020-04-09
Content types with parameters validate wrong request data


jordandobrev:
When the same content type is used with different properties in the requestBody the validation always validates agains the first content type schema.

Example:
```
      requestBody:
        content:
          "application/vnd.company.user+json; type=basic":
            schema:
              $ref: '#/components/schemas/BasicUser'
         "application/vnd.company.user+json; type=internal":
            schema:
              $ref: '#/components/schemas/InternalUser'
```

Atm the content type validation passes but it always validates the request against `BasicUser` instead of `InternalUser` when the request is made with `application/vnd.company.user+json; type=internal`

cdimascio:
This issue is similar to #270. The content matcher is ignoring the boundary. Perhaps, boundary should only be ignored for multipart?

cdimascio:
fixed in `v3.12.0`

Issue #270, opened 2020-04-09
Wildcard content types support


jordandobrev:
Using content types `*.*` does not work atm when specified in requestBody.

```
requestBody:
        content:
          "*/*":
            schema:
```

When route is reached the validation returns `unsupported media type`

cdimascio:
Thanks for the issue @jordandobrev. Would you be up for taking a stab at it? 

Here are some tips:

I believe [this method](https://github.com/cdimascio/express-openapi-validator/blob/53e53d772865d2adedeef89d7702909ebf6e9c8f/src/middlewares/parsers/body.parse.ts#L28) needs to be updated to handle wildcards. That is, it should select a matching wildcard schema if a match is not found. 

You could add a test case for your scenario [here](https://github.com/cdimascio/express-openapi-validator/blob/master/test/request.bodies.ref.spec.ts#L44) and put the openapi bits [here](https://github.com/cdimascio/express-openapi-validator/blob/master/test/resources/request.bodies.ref.yaml#L10)

Note, I've set up Gitpod for this project, so get a full dev environment set up in [one click](https://gitpod.io/#https://github.com/cdimascio/express-openapi-validator). 

Basically, click the link above, once it completes type `npm t` to run the tests. I've also available on our gitter [channel](https://gitter.im/cdimascio-oss/community)

jordandobrev:
Yeah sure I’ll have a go this or next week and will PR it.

cdimascio:
Awesome! Thanks so much.

cdimascio:
available in `v3.12.0`

cdimascio:
@all-contributors add @jordandobrev for code and test

PR #272, opened 2020-04-10
Allow wildcard content types and content types with parameters

jordandobrev:
## Notes
This PR addresses issues #270 and #271 

## Todo
- [x] Add wildcard support for */*
- [x] Add wildcard support for application/*, text/*, etc
- [x] Add support for content types with parameters

### DIFF
--- a/src/middlewares/openapi.response.validator.ts
+++ b/src/middlewares/openapi.response.validator.ts
@@ -106,18 +106,13 @@ export class ResponseValidator {
     const status = statusCode;
     if (status) {
       const statusXX = status.toString()[0] + 'XX';
+      let svalidator;
       if (status in validators) {
-        const svalidator = validators[status];
-        validator = svalidator[contentType];
-        if (!validator) validator = svalidator[Object.keys(svalidator)[0]]; // take first for backwards compatibility
+        svalidator = validators[status];
       } else if (statusXX in validators) {
-        const svalidator = validators[statusXX];
-        validator = svalidator[contentType];
-        if (!validator) validator = svalidator[Object.keys(svalidator)[0]]; // take first for backwards compatibility
+        svalidator = validators[statusXX];
       } else if (validators.default) {
-        const svalidator = validators.default;
-        validator = svalidator[contentType];
-        if (!validator) validator = svalidator[Object.keys(svalidator)[0]]; // take first for backwards compatibility
+        svalidator = validators.default;
       } else {
         throw validationError(
           500,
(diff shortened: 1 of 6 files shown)


test/resources/request.bodies.ref.yaml
```diff
@@ -25,6 +25,19 @@ paths:
             application/hal+json:
               schema:
                 $ref: '#/components/schemas/Test'
+            'application/vnd.api+json; type=one':
+              schema:
+                $ref: '#/components/schemas/Test'
+            'application/vnd.api+json; type=two':
```
jordandobrev, line 31: Always validates agains the first json type found...

src/middlewares/parsers/body.parse.ts
```diff
@@ -44,14 +44,33 @@ export class BodySchemaParser {
       }
 
       if (!content) {
-        const msg =
+        for (const requestContentType of Object.keys(requestBody.content)) {
```
jordandobrev, line 47: Might need to order those in case `*/*` is declared before `application/*`

test/request.bodies.ref.spec.ts
```diff
@@ -54,6 +58,51 @@ describe(packageJson.name, () => {
...
+      })
+      .expect(200)
+      .then(r => {
+        const { body } = r;
+        expect(r.get('content-type')).to.contain('application/ld+json')
+        expect(body).to.have.property('testProperty');
+      });
+  });
+
+  it.only('should return 200 if application/vnd.api+json; type=two request body is validated agains the corrent schema', async () => {
```
jordandobrev, line 90: fails atm due to the was response validation is performed

cdimascio:
@jordandobrev I've updated the response validator to handle the multiple different response types. seems now we need to add similar support for wildcards in the response validator.

test/request.bodies.ref.spec.ts
```diff
@@ -4,7 +4,7 @@ import * as request from 'supertest';
 import { createApp } from './common/app';
 import * as packageJson from '../package.json';
 
-describe(packageJson.name, () => {
+describe.only(packageJson.name, () => {
```
jordandobrev, line 7: MD: drop before merging

test/request.bodies.ref.spec.ts
```diff
@@ -22,8 +22,14 @@ describe(packageJson.name, () => {
         app.post(`${app.basePath}/request_bodies_ref`, (req, res) => {
           if (req.header('accept') && req.header('accept').indexOf('text/plain') > -1) {
             res.type('text').send(req.body);
+          } else if (req.header('accept') && req.header('accept').indexOf('text/html') > -1) {
```
jordandobrev, line 25: Could be simplified by returning the accept header for all of them. Can drop the content type check as well if we do the `req,query.bad_body` first

src/middlewares/openapi.response.validator.ts
```diff
@@ -106,35 +106,61 @@ export class ResponseValidator {
...
       // assume valid
       return;
     }
+    
     if (!body) {
-      throw validationError(500, '.response', 'response body required.');
+      throw validationError(501, '.response', 'response body required.');
+    }
+    
+    // CHECK If Content-Type is validatable
```
cdimascio: This check is causing test errors. 
can you describe this case? I'm sure you added it for a reason. it may be good to add the relevant test case so we can sort out a solution.

cdimascio:
@all-contributors add @jordandobrev for dev and test

Issue #299, opened 2020-05-20
Throw custom Error instead of simple Error object


efabris:
Actually in the request validation process an error is thrown in this way:

`throw validationError(errorCode, req.path, description);`

This approach does not allow to use the power of "istanceof" instruction in the error middleware to detect the error type and add type based logic. The unique way to detect the error type is to check err.status.

A possible solutions could be the use of class and throwing with the "new" operator.

Example:

```
throw validationError(
    400,
    `.query.${q}`,
    `Unknown query parameter '${q}'`,
);
```

Could become:

```
throw new BadRequestError(
    400,
    `.query.${q}`,
    `Unknown query parameter '${q}'`,
);
```

In this way in the error middleware it's possibile to detect the error in this way:

```
if (err instanceof BadRequestError) {
    console.warn(`Caught Bad Request Error for ${req.path}:`, err.fields)
    return res.status(422).json({
      message: 'Validation Failed',
      ....
    })
  }
```

I have seen that the "ono" library is used in the code to get a similar result but it doesn't seem to work with instance checking.

If you want I can work on a possibile implementation of the feature and submit a pull request.

cdimascio:
Available in v3.14.0

PR #301, opened 2020-05-21
Implemented new error throwing

efabris:
Implemented the proposed solutions in the enhancement issue #299 .

### DIFF
--- a/src/middlewares/openapi.multipart.ts
+++ b/src/middlewares/openapi.multipart.ts
@@ -1,11 +1,13 @@
 import { OpenApiContext } from '../framework/openapi.context';
 import { createRequestAjv } from '../framework/ajv';
-import { validationError } from './util';
 import {
   OpenAPIV3,
   OpenApiRequest,
   OpenApiRequestHandler,
   ValidationError,
+  BadRequest,
+  RequestEntityToLarge,
+  InternalServerError,
 } from '../framework/types';
 import { MulterError } from 'multer';
 import ajv = require('ajv');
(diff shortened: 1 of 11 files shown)


src/middlewares/openapi.multipart.ts
```diff
@@ -115,17 +117,24 @@ function error(req: OpenApiRequest, err: Error): ValidationError {
     );
     const unexpected = /LIMIT_UNEXPECTED_FILE/.test(multerError.code);
     const status = payload_too_big ? 413 : !unexpected ? 400 : 500;
-    return validationError(status, req.path, err.message);
+    return payload_too_big
+      ? new RequestEntityToLarge({ path: req.path, message: err.message })
```
cdimascio, line 121: perhaps create a factory helper to automatically throw the proper error object depending on the status code
for example, something like `return HttpError.create({ status, path, message})` and have it returns the correct error instance.
This will help simplify code (like the above) where depending on the status code, a specific error object should be created
e.g. return HttpError.create({ status, path: req.path, message: err.message })

src/middlewares/openapi.security.ts
```diff
@@ -86,11 +98,19 @@ export function security(
...
-      else throw firstError;
+      if (success) {
+        next();
+      } else {
+        throw firstError;
+      }
     } catch (e) {
       const message = e?.error?.message || 'unauthorized';
-      const err = validationError(e.status, path, message);
+      const err =
```
cdimascio, line 108: use the `HttpError.create` helper here to simplify the status based logic

src/middlewares/openapi.multipart.ts
efabris: Great hint. Implemented as requested.

src/middlewares/openapi.security.ts
efabris: Great hint. Implemented as requested.

cdimascio:
@all-contributors add @efabris for code

cdimascio:
@efabris merged. if you have a chance. it would be great to have a couple of tests that validate the `instanceof` the returned `Error` instance is correct.

Issue #292, opened 2020-05-13
Support "fully qualified" operationId that include the handler prefix


mdwheele:
Hello!

☝️ First, thank you so much for this package. I stumbled across this looking for resources that would promote my application's OpenAPI specification as the "source of truth" for Express as well as docgen and this hits all the marks.

I use [Spotlight](https://stoplight.io) to manage my API specification and am able to specify `operationId`, as usual. However, to use this library, I need to modify source directly to specify handler. This works great, but causes a _bit_ of context switching moving between tooling when I really want to stay focused in Stoplight as much as possible (especially when designing)

It might be cool to be able to specify a "fully qualified" `operationId` (or `x-eov-operation-id`) that is relative to the configured `operationHandlers` path. With this proposal, nothing at all would change about the use of existing options. My thoughts would be this might become a "last attempt" before throwing an `Error`.

From the EOV operations example, imagine the following:

```yaml
paths:
  /ping:
    get:
      description: |
        ping then pong!
      operationId: routes/ping/ping
      # This is no longer required as we can specify the handler in our operationId
      # x-eov-operation-handler: routes/ping 
      responses:
        '200':
          description: OK
          content:
            text/plain:
              schema:
                type: string
                example: pong
        default:
          description: unexpected error
          content:
            application/json:
              schema:
                $ref: '#/components/schemas/Error'
```

In my own project, I point `operationHandlers` at a `controllers` directory. So my operationId might look more like `auth.login`, `auth.logout` where my handler is situated as:

```js
# controllers/auth.js

module.exports = {
  login: (request, response) => { /* ... */ },
  logout: (request, response) => { /* ... */ },
}
```

What do you think? I would be _glad_ to work on a PR to draft this up if it would help.

cdimascio:
@mdwheele thanks for the ticket and your interest in the project. I like the idea, though I'm a bit hesitant to support it. here's why...

`operationId` is defined by the OpenAPI specification as a unique string used to identify an operation. Tools e.g. code generators commonly use the value of `operationId` as names for a corresponding method.

As such I don't think I should provide `operationId` with new path-like semantics. It may have downstream consequences. One such consequence that results solely on the path like naming means that clients who consume your spec to generate a client or server stub may end with odd path-like method names.

[...]

PR #312, opened 2020-06-01
Implement the idea of an "Operation Resolver"

mdwheele:
Hello!

💁‍♀️ This PR adds new functionality to allow users to specify a pluggable "Operation Resolver" function that is used to determine the appropriate NodeJS function to call mapped to an OpenAPI Operation.

Related: #292 

In #292, I proposed functionality that would allow the use of "fully qualified `operationId`". An example of the intended usage would be to specify `operationHandlers` to be the path to a set of HTTP controllers (NodeJS modules) and then you would _solely_ rely on `operationId` to map an `Operation` to `RequestHandler` (i.e. `operationId` for a `GET` to `/v1/pets` might be `pets.list`; where a `pets.js` exists at `operationHandlers` and a `list` function is exported from the module).

As I thought about it more, I came to the conclusion that it was highly likely that usage of `operationId` will be implementation-specific and near-impossible to satisfy with a singular strategy. However, if the strategy for `RequestHandler` resolution were made pluggable, then users would be free to plug their own opinion / application-specific needs and as long as they adhere to a standard contract (return a `RequestHandler`), then EOV wouldn't need to know the difference.

So that's what I did. In this PR, we add a new `OpenApiValidator` option called `operationResolver`. It receives a function that accepts two parameters:

1. `basePath: false | string` - This is `options.operationHandlers`, the path to a collection of operation handlers (NodeJS modules)
2. `route: RouteMetadata` - Information about the current route as `installOperationHandlers` is executed.

Once extracted, we supply a default implementation (the current strategy utilizing `operationId` and two extended attributes: `x-eov-operation-id` and `x-eov-operation-handler`). If a user specifies nothing, then they get _exactly_ the behaviour as it exists today.

I've added a fifth example application that makes use of this new feature to implement the original feature request I made in #292:

```js
new OpenApiValidator({
  apiSpec,
  validateResponses: true,
  operationHandlers: path.join(__dirname, 'routes'),
  // Provide a function responsible for resolving an Express RequestHandler 
  // function from the current OpenAPI Route object.
  operationResolver: function (basePath, route) {
    // Process operationId and assume {controller}.{method} format.
    [controller, method] = route.schema['operationId'].split('.')

    // Build absolute path to controller module 
    const modulePath = path.join(basePath, controller);

    // I decide _not_ to catch errors here, but one could do if they wished
    // The default error handling when hitting this was clear enough for me.
    const handler = require(modulePath)

    // My own error checking to verify that the export exists on the module.
    if (handler[method] === undefined) {
      throw new Error(`Couldn't find a [${method}] function in ${modulePath} when trying to route [${route.method} ${route.expressRoute}].`)
    }

    // Finally, return the handler function
    return handler[method]
  }
})
```

Nothing changes about how I export functions from my operation handler and nothing really changes about how the rest of this library works after resolution occurs. The only change I end up making is _this_ configuration AND my desired `operationId: pets.list`-style usage.

I think there are some improvements that can be made to how an "operation resolver" is typed. Currently, I typed it as a simple function. However, I could totally see creating a new type `OperationResolver` and getting strict about it's parameters and return type. Please let me know if you'd like me to do this.

Lastly, if we move forward with this idea, it may be reasonable to use this approach to collect a set of supported resolvers. It's not a requirement, but it would help grease adoption for folks and also demonstrate levels of flexibility offered by the library.

### DIFF
--- a/src/framework/types.ts
+++ b/src/framework/types.ts
@@ -62,6 +62,7 @@ export interface OpenApiValidatorOpts {
     mode: 'bundle' | 'dereference';
   };
   operationHandlers?: false | string;
+  operationResolver?: Function;
   validateFormats?: false | 'fast' | 'full';
 }
 
--- a/src/framework/types.ts
+++ b/src/framework/types.ts
@@ -47,6 +47,11 @@ export type ValidateSecurityOpts = {
   handlers?: SecurityHandlers;
 };
 
+export type OperationHandlerOptions = {
+  basePath: string,
+  resolver: Function
+}
+
 export interface OpenApiValidatorOpts {
   apiSpec: OpenAPIV3.Document | string;
   validateResponses?: boolean | ValidateResponseOpts;
@@ -61,8 +66,7 @@ export interface OpenApiValidatorOpts {
   $refParser?: {
     mode: 'bundle' | 'dereference';
   };
-  operationHandlers?: false | string;
-  operationResolver?: Function;
+  operationHandlers?: false | string | OperationHandlerOptions;
   validateFormats?: false | 'fast' | 'full';
 }
 
--- a/src/index.ts
+++ b/src/index.ts
@@ -16,6 +16,8 @@ import {
 import { deprecationWarning } from './middlewares/util';
 import * as path from 'path';
 import { BasePath } from './framework/base.path';
+import { defaultResolver } from './resolvers';
+import { OperationHandlerOptions } from './framework/types'
 
 export {
   InternalServerError,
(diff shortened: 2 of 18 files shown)


src/framework/types.ts
```diff
@@ -62,6 +62,7 @@ export interface OpenApiValidatorOpts {
     mode: 'bundle' | 'dereference';
   };
   operationHandlers?: false | string;
+  operationResolver?: Function;
```
cdimascio, line 65: i'm wondering, if `operationHandlers` and `operationResolvers` might be collapsed into a single property.

something like

```javascript
    operationHandlers: {
        path,
        resolver
    }
```

To maintain backward compatibility, the `operationsHandlers` would continue to support `false` and `string`, but would now also support a new type e.g. `OperationHandlerOptions` with properties `path` and `resolver`

Something like the above. Open to suggestions.

src/index.ts
```diff
@@ -44,6 +45,45 @@ export class OpenApiValidator {
     if (options.operationHandlers == null) options.operationHandlers = false;
     if (options.validateFormats == null) options.validateFormats = 'fast';
 
+    if (options.operationResolver == null) {
+      options.operationResolver = (handlersPath: false | string, route: RouteMetadata) => {
```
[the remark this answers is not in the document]
mdwheele: Agreed 100%. I've created a new module at `src/resolvers.ts` that exports the default resolver as well as the resolver I wrote up for the example. I think this might be where we can ship supported resolvers for different use-cases. I'm going to add some more thorough error-checking to the resolver I wrote in a bit, but wanted to go ahead and push what I had up.

src/framework/types.ts
mdwheele: Totally! I just pushed some in-progress work that collapses all this down. I've got some failing tests regarding coverage on the new `resolvers.ts` to take care of. We'll also likely want some tests that verify the logic used to map the public API for `operationHandlers` to `OperationHandlerOptions` internally.

mdwheele:
> Thinking about this mainly because the sample resolver you created (I believe) may be very useful to folks. In any case, this is definitely not required, but wanted to mention it in case you have ideas. (not sure this is a good or bad idea ;) just putting it out there)

The nice thing about this direction is folks are free to implement whatever resolver strategy they want; as long as it returns a `RequestHandler` at the end. With that said, I agree with you that providing a few alternative resolvers is a good idea. I don't have a strong opinion whether that's in a separate package, built-in here or in an end-user's application. 

I think having the current default resolver in addition to two supported resolvers with good documentation would be more than a great start for users:

1. "Module Path Resolver" (naming is hard): This resolver uses a convention that there will be a flat directory at some provided base path with a bunch of modules that export `RequestHandler` functions. `operationId` follows the template of `{moduleName}.{functionName}`. Deeply nested `operationId` may or may not be supported. I'd say that this resolver is really more of an example on how you might write your own. It may be too simplistic for some folks that want a more deeply nested handlers directory, but it provides excellent guidance.
2. "RESTful Resolver": This resolver might use a REST-based convention for determining which function to call when handling a request. In this use-case, the user may not even provide an `operationId` at all so you get this effect of "automatic routing" which some might like. The function used to handle Express requests would be inferred from the path and method of the Operation object in the schema. For example, a request to `/v1/pets/4` would map to `pets.show` and the expectation would be that you still have to provide a handlers path, but you don't write `operationId` into your schema any more. It's just going to expect `/path/to/handlers/pets.js` to export a `show` function.

Mostly just some thoughts/ideas. I still need to write up some documentation and examples explaining this option a little better. If you want to take another look at things as they stand, I can adjust docs as we go.

src/framework/types.ts
cdimascio: awesome. let me know how it goes

cdimascio:
Cool. this looks good. I'm going to play around with it this weekend.
I'm still unsure on the documentation.

What you've written up is certainly useful, but definitely aimed at a power user. I wonder if it should include something for the standard user.

In what you've implemented, you've provided, in some respects, a simpler way to resolve modules. Arguably simpler than what exists today, since all that is needed is `operationId`. This may be beneficial to folks who don't care about code-gen tooling (and documentation tooling). 

Also, this also overlaps quite a bit with the existing mechanism. I'm basically concerned about confusion. 

Anyway, I'll let more on this over the next day or so.
Feel free to continue to make updates if you see fit

Thanks for all your work! Much appreciated!

cdimascio:
i've made some modifications to the README. 
The big change was to move the larger example doc from the main README and add it to example 5's README. 
I also reference example 5's readme the `operationHandlers` section of the main README

cdimascio:
@all-contributors add @mdwheele for code, doc, and test

cdimascio:
@mdwheele thanks for you work on this. It is available in `v3.16.0`

PR #322, opened 2020-06-17
Update api.yaml for /pets route. Both x-eov-operationId and operation…

tmcarmichael:
The route /pets was updated and exports, 'pets#list'. The 'x-eov-operation-id' param has this updated value, however 'operationId' has 'findPets' which cannot be found in the pets route. Further, only the 'x-eov-operation-id' is necessary as it overrides 'operationId' as noted in comment on line 17: https://github.com/cdimascio/express-openapi-validator/blob/master/examples/3-eov-operations/api.yaml#L17. 

**_Summary_**: this line can be removed 
https://github.com/cdimascio/express-openapi-validator/blob/2f0c206abb0490909bf0ea48d5cdde204544201c/examples/3-eov-operations/api.yaml#L38

### DIFF
--- a/examples/3-eov-operations/api.yaml
+++ b/examples/3-eov-operations/api.yaml
@@ -14,7 +14,7 @@ paths:
     get:
       description: |
         ping then pong!
-      # use operationId, (x-eov-operation-id is optional. it can be used in place of operationId or to override it)
+      # operationId is used in /ping path (x-eov-operation-id is optional). x-eov-operation-id can be used in place of operationId. If both x-eov-operation-id and operationId are present, x-eov-operation-id will override operationId.
       operationId: ping 
       x-eov-operation-handler: routes/ping 
       responses:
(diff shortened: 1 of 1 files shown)


cdimascio:
Rather than removing `operationId`, perhaps a comment should be introduced. The reason `operationId` is present in the example is to illustrate that if `operationId` is already specified or necessary, we we can provided the eov extension (This is useful for tools e.g. code generators that use  the `operationId` value to generate method names)
This allows us to both use `operationId` for that use case, yet still name our js methods differently

tmcarmichael:
Alright, that makes sense, I'll add a comment that summarizes the info above for the eov examples.

tmcarmichael:
Updated PR using original example 3 EOV api yaml with two comments detailing the use of 'operationId' and 'x-eov-operation-id'. As well as backwards compatibility with code generator tools by keeping 'operationId'.

I'm enjoying EOV, if there's some way I can contribute let me know :).

tmcarmichael:
Update: squashed the commits down to 1, and updated the commit message to reflect the updated comment in example 3 eov-operations.

cdimascio:
@all-contributors add @tmcarmichael for documentation

PR #339, opened 2020-07-22
unused deps cookie-parser in examples and updated readme

xtrycatchx:
- Removed unused cookie-parser in package.json as well as in the sample code itself 
- Removed the non-existing `public` path exposed as static by the sample apps
- Updated the README to reflect these small changes


cdimascio:
@all-contributors add @xtrycatchx for code

PR #351, opened 2020-08-14
implement as standard connect  middleware

cdimascio:
#352


Envek:
I found this repo today and decided to try v4.0.0.alpha1 because I liked middleware usage much more than promise-based one. And it works very well for my simple case. Thank you for express-openapi-validator!

cdimascio:
@envek awesome and a huge thank you for posting your feedback. Glad to hear it's working for your use case. Please post any issues that you run into. Thanks again!

mahnunchik:
V4 doesn't works with `express.Router()`.

Works as expected:
```js
const app = express();

const api = express();
api.use(OpenApiValidator.middleware(opts));

app.use(api);
```

Doesn't works - every route is 404 error:
```js
const app = express();

const api = express.Router();
api.use(OpenApiValidator.middleware(opts));

app.use('/api/', api);
```

Both option uses the same spec with:

```yml
servers:
  - url: /api/
```

cdimascio:
@mahnunchik thank you for submitting this. I will definitely have a look

cdimascio:
i modified `examples/3-eov-operations, and used

and replaced `app.js` with the following contents

```javascript
const express = require('express');
const path = require('path');
const bodyParser = require('body-parser');
const logger = require('morgan');
const http = require('http');
const OpenApiValidator = require('express-openapi-validator');

const port = 3000;
const app = express();
const apiSpec = path.join(__dirname, 'api.yaml');

app.use(bodyParser.urlencoded({ extended: false }));
app.use(bodyParser.text());
app.use(bodyParser.json());

app.use(logger('dev'));
app.use('/spec', express.static(apiSpec));

// 1. Create a router
const api = express.Router();  

// 2. Use oav middleware with router
api.use(
  OpenApiValidator.middleware({
    apiSpec,
    validateResponses: true,
    operationHandlers: path.join(__dirname),
  }),
);

// use the router on the app
app.use('/v1', api);   

app.use((err, req, res, next) => {
  // format errors
  res.status(err.status || 500).json({
    message: err.message,
    errors: err.errors,
  });
});

http.createServer(app).listen(port);
console.log(`Listening on port ${port}`);

module.exports = app;
```

I then curl'ed an endpoint. It is works 
```shell
curl 'localhost:3000/v1/pets?type=cat&limit=' 
{"message":"Empty value found for query parameter 'limit'","errors":[{"path":".query.limit","message":"Empty value found for query parameter 'limit'"}]}%   
```

Would you mind testing out something similar.

Also, please create a github repo with your failing example. i'll have a look.

Do note that you will need the body parser middleware...though likely not the cause of the `404`

cdimascio:
@mahnunchik  see comment above

mahnunchik:
Hi @cdimascio 

The root of the issue is due to async nature of initialization of `OpenApiValidator`.

My code suddenly includes the following 404 handler:

```js
api.use(
  OpenApiValidator.middleware({
    apiSpec,
    validateResponses: true,
    operationHandlers: path.join(__dirname),
  }),
);

app.use('/v1', api);   

// actually any handler after API handler
app.use((req, res, next) => {
  res.status(404).send({ error: 'Page not found.' });
});
```

In this example `*` -> `404` handler will be added before any routes from `OpenApiValidator` due to async spec loader.

I think that `OpenApiValidator.middleware` should loads synchronously to mitigate unpredictable issues related to order of route handlers.

cdimascio:
@mahnunchik try `v4.0.0-alpha.4`, `operationHandlers` it should work when using `app.use(api)`. note that there is still some work to do to handle the route prefix e.g. `app.use('/v1', api)`

cdimascio:
@mahnunchik give `v4.0.0-alpha.5` a try. it should work when specifying paths now e.g. `app.use('/v1', api)`

mahnunchik:
@cdimascio thanks! `v4.0.0-alpha.5` works with mounted Router as expected.

cdimascio:
Excellent! Thanks a lot for trying out the alpha. Super great to have your help

Issue #359, opened 2020-09-04
readOnly + required not handled properly


dchekanov:
readOnly + required combination is supposed to be working (#145), but it doesn't in my case. 

I traced the issue to the [`cleanseContentSchema`](https://github.com/cdimascio/express-openapi-validator/blob/0447c47996fb77b5d4e29917e50ef4327b6786be/src/middlewares/parsers/body.parse.ts#L85) function.

contentType in my requests (made with axios) equals to this:

```javascript
ContentType {
  contentType: 'application/json;charset=utf-8',
  mediaType: 'application/json',
  charSet: 'charset=utf-8',
  withoutBoundary: 'application/json;charset=utf-8'
}
```

And here's the requestBody:

```javascript
{
  description: 'StorePlatform properties',
  content: { 'application/json': { schema: [Object] } }
}
```

[The function](https://github.com/cdimascio/express-openapi-validator/blob/0447c47996fb77b5d4e29917e50ef4327b6786be/src/middlewares/parsers/body.parse.ts#L90-L91) looks for `requestBody.content[contentType.withoutBoundary]`, which translates to `requestBody.content['application/json;charset=utf-8']`, which does not exist, so the fix for the combination is not applied.

It works if I replace `requestBody.content[contentType.withoutBoundary]` with `requestBody.content[contentType.mediaType]`. Not sure if that's the correct way to deal with the issue. Would appreciate if someone could take a look into it.

cdimascio:
@dchekanov currently, it requires a match on the content type + charset. it does not match on boundary. one way to solve is to update your schema to include both `application/json` and `application/json;charset=utf-8`.

i think it's also worth considering whether `charset` (like boundary) should be ignored in the test, thus only matching only matching on `media type`

cdimascio:
thanks for the fix, @dchekanov. its in `v3.16.14`

dchekanov:
Thank you @cdimascio for responding and releasing a new version so quickly!

PR #362, opened 2020-09-06
Handle readOnly + required when Content-Type includes charset

dchekanov:
See #359. This version relies on the `content` object already matched on the previous stage of processing, supporting Content -Type with and without a charset.


cdimascio (approved):
this looks good. thanks @dchekanov

cdimascio:
@all-contributors add @dchekanov for code and tests

PR #374, opened 2020-09-24
WIP: Example operation resolver

michaelmelody91:
Following on from https://github.com/cdimascio/express-openapi-validator/pull/366, I've pieced together this PR as a proof of concept in how an operation resolver may return examples defined in the API definition.

This PR so far fetches the `example` defined on the schema object. This could be further expaneded to return the singular example or one of the multiple examples defined on the `MediaTypeObject`.

Using the `apiDoc` param introduced in #366, I'm derefrencing the route object. To enable this I've introduced [json-schema-deref-sync](https://www.npmjs.com/package/json-schema-deref-sync) (this is not ideal and something I'd very much like to avoid.. I've struggled with introducing a sycronous implementation for the exsting **json-schema-ref-parser** library being used. See https://github.com/APIDevTools/json-schema-ref-parser/issues/14#issuecomment-698342351).

**Derefrencing route using apiDoc**
```
      const dereferencedSchema = deref({route, ...apiDoc});
      const responses = dereferencedSchema.route.schema.responses;
```

I had also attempted to replicate something like https://github.com/cdimascio/express-openapi-validator/blob/master/src/framework/openapi.spec.loader.ts#L55-L68 - but was unable to get this working for me so far.

Question for @cdimascio, does this sort of resolver belong in the main project or is this something that should be for the user to worry about? I think introducing a mock server element which is driven from the spec could be something really powerful but also feel that trying to cover the multiple different styles of writing an API (using refs, example vs examples, example on the media object or the schema etc) could prove difficult when trying to write a catch all resolver.

## TODO

- Tests
- Handle multiple examples


cdimascio:
thanks @michaelmelody91 i'll have a look at this PR soon.
in the meantime, it looks like there are a couple merge conflicts to resolve

PR #381, opened 2020-09-30
fix #380: allow any string to be servers variables

### LINKED ISSUE #380: Why do we validate the server variables? There is no such limitation in OpenAPI 3 specification
## Problem description

if we have a openapi 3.0 specification like this:

```yaml
servers:
  - url: https://api.example.com/{basePath}
    description: Production server (uses live data)
    variables:
        basePath:
            default: "v1/path-a/path-b/pathC

```

It will report error that the `basePath` doesn't match the pattern "[^\/#\?]+?"

## Discussion

This pattern check comes from the package `path-to-regexp`.

But I don't think we need such constrains. There is no such limitation in OpenAPI 3 specification

According to https://swagger.io/docs/specification/api-host-and-base-path/, it says:

> Variables can have arbitrary values, or may be restricted to an enum. 

image

xg1990:
According to the OpenAPI 3.0 specification, we should be able to use any variables in servers. More comments can be found in this issue: https://github.com/cdimascio/express-openapi-validator/issues/380

This change is based on the latest documentation of [path-to-regexp](https://github.com/pillarjs/path-to-regexp) that we can specify the pattern of each path parameters

### DIFF
--- a/src/framework/base.path.ts
+++ b/src/framework/base.path.ts
@@ -20,7 +20,7 @@ export class BasePath {
     let urlPath = this.findUrlPath(server.url);
     if (/{\w+}/.test(urlPath)) {
       // has variable that we need to check out
-      urlPath = urlPath.replace(/{(\w+)}/g, (substring, p1) => `:${p1}`);
+      urlPath = urlPath.replace(/{(\w+)}/g, (substring, p1) => `:${p1}(.*)`);
     }
     this.path = urlPath;
     for (const variable in server.variables) {


xg1990:
This might also fix #115

cdimascio:
i see. so looks like the current regex is not allowing you to have `/`s in the variable name. am i understanding this correctly?

with your new change, we'll end up with simple paths, looking similar to the following
```shell
// given the path 'http://example.com/test/{me}/hello/{there}'

> urlPath.replace(/{(\w+)}/g, (substring, p1) => `:${p1}(.*)`)

// produces
'http://example.com/test/:me(.*)/hello/:there(.*)'
```

This seems reasonable. The new matcher will have the same behavior as the previous logic, however we can now match the `/`s

Assuming, I'm understanding the issue correctly, your change looks reasonable to me

xg1990:
Hi @cdimascio ,

Thanks for this response.

> i see. so looks like the current regex is not allowing you to have /s in the variable name. am i understanding this correctly?

Yes, to be more precise, it's `/` in `variable value` 

And your understanding is right, according to `path-to-regexp` 's documentation, if we have `/:there(.*)` then it will allow anything in the path

cdimascio (approved):

cdimascio:
@all-contributors add @xg1990 for code

cdimascio:
@xg1990 your fix is in `v4.0.6`

PR #383, opened 2020-10-01
Allow JSON Response to Return Boolean `false` Value

DomParfitt:
Currently the implementation will fail to validate responses that return a simple JSON boolean value in the case where the value returned is `false`.

This is due to a check in the response validator to see if the body of a response exists and returns a `500` response if it does not, which currently uses a shorthand check for the existence of the response body. As a result if the response body is set to a simple `false` value it is treated as if the body does not exist.

### DIFF
--- a/src/middlewares/openapi.response.validator.ts
+++ b/src/middlewares/openapi.response.validator.ts
@@ -54,8 +54,8 @@ export class ResponseValidator {
         const accepts: [string] = contentType
           ? [contentType]
           : accept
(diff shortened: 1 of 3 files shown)


cdimascio (approved):
thanks @DomParfitt

cdimascio:
@all-contributors add @DomParfitt for code and test

cdimascio:
@DomParfitt your fix is in `v4.0.6`

Issue #268, opened 2020-04-02
Any work-around for ajv's undesired number coercing rules?


bluenote10:
**Is your feature request related to a problem? Please describe.**

Since this library is using ajv under the hood, the issue is probably a fundamental problem in ajv, which I have reported [here](https://github.com/epoberezkin/ajv/issues/1182). I'm not sure if this will be actioned on ajv side, so I'm wondering if a work-around is feasible in terms of the hooks in the middleware.

In short: Coercing the value `null` for a number currently maps to 0. I'm working on an API where request data may contain NaN/Infs basically everywhere, and the roundtrip `JSON.stringify` => `JSON.parse()`=> `validate` maps them to 0, leading to bugs everywhere.

**Describe the solution you'd like**

Would it be possible to pass in special validation hooks, i.e., a callback system that would allow to handle the case "received value `null` for type `number`, how to proceed?". I have no idea if an approach like that would be feasible.

**Describe alternatives you've considered**

You are the experts, are there better ways to work-around ajv's behavior?

**Additional context**

See linked issue above.

cdimascio:
One possible workaround is to add `nullable: true`. 

```yaml
    TestObject:
      type: object
      properties:
        number:
          type: integer
          nullable: true
```

bluenote10:
As I have tried to explain in the other ticket already, this is not really a solution.

Maybe it helps to explain the context. The API is in the context of numerical data science, where data gets pushed from Python clients. In the data science community it is a standard to represent missing numbers as `NaN` (Numpy/Pandas follow that rule). The API consists of maybe 15 types all containing mainly `number` fields. The total number of `number` fields is approx 100. Now basically all of these fields can contain a `NaN` on sender side (Python), and all of them should be transferable 1-to-1 into the backend. 

[...]

Issue #249, opened 2020-02-27
Request body property is coerced e.g. ("123" coerced to123)


AKGP:
Hi,  I am using openapi: 3.0.2 with express-openAPI-validator ^2.0.0
I am getting the following problem:
When I am using 
```
new OpenApiValidator({
      apiSpec: this.openApiPath,
      validateResponses: false,
      **validateRequests:** true,
    }).install(this.app);
```

its changing the integer values to string
like 
```
body: { a: 1}
```
and I am getting 

```
body: {a:"1"}
```

Please fix this problem as I am getting validation issue due to this

cdimascio:
@AKGP, can u share your spec. I wonder if the property `a` is specified as a string and not a number

AKGP:
```
properties:
monthlyIncome:
          type: string
          pattern: ^[0-9]+\.?[0-9]*$
```
i am using this but when I put the value as 100 in request body I found it got changed to "100"after validation instead it show throw an error of type mismatch.

cdimascio:
Because you spec specifies monthly income as a string. AJV will coerce the value according to its [coercion rules](https://github.com/epoberezkin/ajv/blob/master/COERCION.md). In this case, `100`can be coerced to the correct type, thus its coerced from the number `100` to expected type string, `"100"`. 

This is currently the expected behavior,

chittolina:
@cdimascio First of all - really loved this lib. ty so much.

Now, I have a question. Why would the values be coerced *before* the validation? If that's the expected behavior, basically there's no way to use this to return errors in case the user inputs a `number` where we expect a `string`? 

I'm having the same issue right now. This is my definition:

```
 PropertySearchInput:
     type: object
     required:
         - deliveryLine
     properties:
         zip:
             type: string
             pattern: '^\d{5}$'
         deliveryLine:
             type: number
```

Using this request:

```
{"userID": "123", "deliveryLine": "445"}
```

Works fine, which, IMO, doesn't seem right. If I specify that my property is a `number`, why would a `string` be allowed?

Is it possible to add a flag to enable/disable it?

cdimascio:
[...]

Issue #243, opened 2020-02-25
coerceTypes is all or nothing


jtinbergen:
When you do not want coerceTypes to correct malformed payloads sent into your API you MUST mark your openAPI parameters as String or no API call will succeed.

/test/{number}/detail/{number2}

This call will fail since parameters are always strings (you do not have a choice!).

Ideally you would want to always coerce parameters when possible (since you don't have a choice in sending them as string) and let the body/payload optionally be coerced if desired.

cdimascio:
thanks @jtinbergen. good point. we'll look into it

cdimascio:
this sounds like the same problem as described in https://github.com/cdimascio/express-openapi-validator/issues/249#issuecomment-636499628. thoughts?

cdimascio:
@jtinbergen i've put together [PR](#387 ) which attempts to solve these issues via a behavioral change.
- type coercion will will continue to apply to query, path,and header params
- type coercion will not apply to request bodies.

as a result, the `coerceTypes` option becomes a bit ambiguous. i plan to deprecate the property as it's not clear what its behavior should be. all or nothing is not desired. there is an argument for two options, but believe, the behavior that makes the most sense is what i've outlined above. with this behavior, i don't know that a type coercion option is needed.

thoughts?

cdimascio:
fixed in `v4.1.0`

PR #387, opened 2020-10-04
do not coerce request body property values

cdimascio:
This PR remove type coercion from request bodies.
This should fix the following issues: 
- Request body property is coerced e.g. ("123" coerced to123) #249
- Any work-around for ajv's undesired number coercing rules? https://github.com/cdimascio/express-openapi-validator/issues/268. 
   **Note**: This fix will ensure that `null`s passed as numbers (by default) fail validation. 
                    Issue #268 would like support for round-trip `NaN`s which this does not solve
- coerceTypes is all or nothing #243

### DIFF
--- a/README.md
+++ b/README.md
@@ -696,7 +696,7 @@ Specifies the options to passthrough to multer. express-openapi-validator uses m
   }
   ```
 
-### ▪️ coerceTypes (optional)
+### ▪️ coerceTypes (optional) - _deprecated_
 
 Determines whether the validator should coerce value types to match the type defined in the OpenAPI spec.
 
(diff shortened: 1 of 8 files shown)


MadMango:
I just found this issue after upgrading the library and finding that it's broken because I'm sending numbers as strings to our APIs.

There are good reasons to have coercion enabled as explained [here](https://github.com/ajv-validator/ajv/blob/master/docs/validation.md#coercing-data-types)

@cdimascio  What was the rationale behind removing the option completely? The problem you references could have been solved by just setting coerceTypes to false.

cdimascio:
Thanks for your input. Here is the rationale. Happy to open the discussion.
https://github.com/cdimascio/express-openapi-validator/issues/396#issuecomment-707840224

PR #393, opened 2020-10-08
implement and/or logic in security validator

balazssoltesz:


cdimascio:
@balazssoltesz Thanks a lot for the PR!

There appears to be one test that needs attention
you can run them locally via `npm t`

``` 
1) security.handlers
       should return 200 if api_key or anonymous and no api key is supplied:
     Error: expected 200 "OK", got 401 "Unauthorized"
      at Test._assertStatus (node_modules/supertest/lib/test.js:268:12)
      at Test._assertFunction (node_modules/supertest/lib/test.js:283:11)
      at Test.assert (node_modules/supertest/lib/test.js:173:18)
      at Server.localAssert (node_modules/supertest/lib/test.js:131:12)
      at emitCloseNT (net.js:1658:8)
      at processTicksAndRejections (internal/process/task_queues.js:79:21)
```

in addition to investigating this error, it will be good to add additional tests to exercise the new and/or capabilities

thanks so much!

balazssoltesz:
i not rly understand this. why should return 200 if not api_key supplied? That is the point, we should return 401 when any of the security check is failed (when AND logic is used).

cdimascio:
@balazssoltesz openapi allows one to declare anonymous access explicitly. this particular test, checks that if its is declared that ApiKey or anonymous access is enabled for an endpoint, that it should return 200 if either the request contains the apikey or does not i.e  anonymous

Here is the spec that declares this. Note that `{}` indicates anonymous:

https://github.com/cdimascio/express-openapi-validator/blob/master/test/resources/security.yaml#L27-L37

```
  /api_key_or_anonymous:
    get:
      security:
        # {} means anonyous or no security - see https://github.com/OAI/OpenAPI-Specification/issues/14
        - {}
        - ApiKeyAuth: []
      responses:
        "200":
          description: OK
        "401":
          description: unauthorized
```

all in all, its possible the new code is not considering the `{}` case

cdimascio:
note that in my example above (which the test exercises), it states A **OR** B where A is `{}` (anonymous) and B is `ApiKey`.  In this case, if no authentication is provided, it should return success as  it satisfies the security conditions

cdimascio:
@balazssoltesz Would u mind adding a test or two.  
This is fantastic! Thanks so much for your work on this.

cdimascio:
@all-contributors add @balazssoltesz for code and test

PR #401, opened 2020-10-20
Move @types/multer to dependencies

dprgarner:
Types from `@types/multer` are being used in the exported `framework/types.d.ts` file. This breaks the TypeScript build of any project that uses this library but doesn't have `@types/multer` installed. Multer isn't a peerDependency of this library, so it shouldn't be a requirement for projects using express-openapi-validator to install it separately. It should be included as a dependency, and not a devDependency. 

There could potentially be an issue with `@types/express` being used in `framework/types.d.ts` as well, but it's likely that any TypeScript project using this library will already have express types installed separately.

### DIFF
--- a/package.json
+++ b/package.json
@@ -31,6 +31,7 @@
   "author": "Carmine DiMascio <cdimascio@gmail.com>",
   "license": "MIT",
   "dependencies": {
+    "@types/multer": "^1.4.4",
     "ajv": "^6.12.5",
     "content-type": "^1.0.4",
     "js-yaml": "^3.14.0",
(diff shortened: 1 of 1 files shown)


cdimascio (approved):
Thanks @dprgarner!

cdimascio:
@all-contributors add @dprgarner for code

pscanf:
@cdimascio I saw you ~reverted this PR [soon after](https://github.com/cdimascio/express-openapi-validator/pull/407/commits/cce2713890a1c6f3f08b88f7afdfdd347c5b69af), were there any problems with moving `@types/multer` to dependencies?

(I noticed as updating to the latest version of `express-openapi-validator` once again gives me typescript errors regarding multer).

P.S. - since this is the first issue I open here, I'll take the chance to thank you for this amazing project! Great approach for building REST APIs and great DX of the library!

cdimascio:
@dprgarner apologies, my mistake. This can go back to dependencies

PR #419, opened 2020-10-26
(fix) #415 Cannot read property 'push' of undefined bug

### LINKED ISSUE #415: unhandledRejection: TypeError: Cannot read property 'push' of undefined
**Describe the bug**
Regression appears after #414

```javascript
// request.schema.preprocessor.js

preprocessPathLevelParameters(pathItemKey, pathItem) {
        var _a, _b, _c;
        const parameters = (_a = pathItem.parameters) !== null && _a !== void 0 ? _a : [];
        if (parameters.length === 0)
            return;
        const v = pathItem[pathItemKey];
        if (v === parameters)
            return;
        const ref = (_b = v === null || v === void 0 ? void 0 : v.parmeters) === null || _b === void 0 ? void 0 : _b.$ref;
        const operationParameters = (ref ? (_c = this.ajv.getSchema(ref)) === null || _c === void 0 ? void 0 : _c.schema : v.parameters);
        for (const param of parameters) {
            operationParameters.push(param); // <-- this place
        }
    }
```

**Actual behavior** 
Error throws if `parameters` property exists in `path` but doesn't exists in `pathItem`

**Expected behavior**
`pathItem` can be without `parameters` property

**Examples and context**
```json
{
  "openapi": "3.0.0",
  "paths": {
    "/product/{productId}/prices": {
      "parameters": [{
        "in": "path",
        "name": "productId",
        "required": true,
        "schema": {
          "$ref": "#/components/schemas/DualID"
        }
      }],
      "patch": {
        "summary": "Update price values.",
        "tags": ["Price"],
        "requestBody": {
          "required": true,
          "content": {
            "application/json": {
              "schema": {
              }
            }
          }
        }
      }
    }
  },
  "schemas": {
    "UniqID": {
      "type": "string",
      "pattern": "^([a-f0-9]){24}$",
      "description": "Hexadecimal unique identifier",
      "example": "5dee464910450609ca7ea2d7"
    },

    "InternalID": {
      "type": "string",
      "pattern": "^([A-Z0-9]){5,6}$",
      "description": "Short identifier that uniq for certain entity",
      "example": "XQ02F"
    },

    "DualID": {
      "oneOf": [{
          "$ref": "#/components/schemas/UniqID"
        },
        {
          "$ref": "#/components/schemas/InternalID"
        }
      ],
      "description": "Dual unique identifier, accepts long and short identifiers transparently"
    }
  }
}
```

ex1st:
Hi @cdimascio, after that fix the validator works fine.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,6 +1,6 @@
 {
   "name": "express-openapi-validator",
-  "version": "4.3.2",
+  "version": "4.3.3",
   "lockfileVersion": 1,
   "requires": true,
   "dependencies": {
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "express-openapi-validator",
-  "version": "4.3.2",
+  "version": "4.3.3",
   "description": "Automatically validate API requests and responses with OpenAPI 3 and Express.",
   "main": "dist/index.js",
   "scripts": {


src/middlewares/parsers/request.schema.preprocessor.ts
```diff
@@ -61,6 +61,10 @@ export class RequestSchemaPreprocessor {
     if (v === parameters) return;
     const ref = v?.parameters?.$ref;
 
+    if (!this.ajv.getSchema(ref)?.schema) {
```
cdimascio, line 64: if say `op = this.ajv.getSchema(ref)?.schema` returns a schema, you'll want to simiarly do `op.parameters = op.parameters || [];`

assuming you make that change, you can remove this block entirely (which will be nice. its ugly)
```
   const operationParameters = <
      Array<OpenAPIV3.ParameterObject | OpenAPIV3.ReferenceObject>	      Array<OpenAPIV3.ParameterObject | OpenAPIV3.ReferenceObject>
    >(ref ? this.ajv.getSchema(ref)?.schema : v.parameters);	    >(ref ? this.ajv.getSchema(ref)?.schema : v.parameters);
```

then change

```
    for (const param of parameters) {
      operationParameters.push(param);
    }
```
to
```
    for (const param of parameters) {
      v.parameters.push(param);
    }
```

src/middlewares/parsers/request.schema.preprocessor.ts
```diff
@@ -57,16 +57,18 @@ export class RequestSchemaPreprocessor {
...
     if (v === parameters) return;
     const ref = v?.parameters?.$ref;
 
-    const operationParameters = <
-      Array<OpenAPIV3.ParameterObject | OpenAPIV3.ReferenceObject>
-    >(ref ? this.ajv.getSchema(ref)?.schema : v.parameters);
+    const op = ref && this.ajv.getSchema(ref)?.schema;
+    if (op)
+      v = op;
+    else
```
cdimascio, line 67: you don't need this else. 

basically, `v` is either a reference to an operation object or the operation object itself.
if its a ref, you dereference via ref && this.ajv.getSchema(ref)?.schema to get the operation object, otherwise you already have it

cdimascio: in both cases you do  `v.parameters = v.parameters || [];`

Issue #426, opened 2020-10-27
4.3.x Custom resolvers are incompatible with `options.apiSpec` set to path.


aviskase:
Related to #416 

**Describe the bug**
After upgrading to 4.3.x custom resolvers are incompatible with `options.apiSpec` set to path.

In 4.2.3 custom resolver was working with:
```
resolver: (basePath: string, route) => {
 const functionName = route.schema['operationId'];
 ...

```

In 4.3.x custom resolver had to be change to different approach because there is no longer `.schema` property in the `route`:

```
resolver: (basePath: string, route, apiDoc) => {
  const pathKey = route.openApiRoute.substring(route.basePath.length);
  const schema = apiDoc.paths[pathKey][route.method.toLowerCase()];
  const functionName = schema['operationId'];
  ...
```

But it relies on  `installOperationHandlers` to send `apiDoc` always as an object, whereas per line https://github.com/cdimascio/express-openapi-validator/blob/ed7af2b02fa45c069c5bdedf5ccf1eff914e3610/src/openapi.validator.ts#L321 it sends whatever was defined in options (in my case, path to API description doc).

aviskase:
I found that `apiDoc` argument was introduced in #366 in version 3.17.x , so I think the are two problems here:

- for some reason, `route.schema` is now undefined
- `apiDoc` is better to be loaded into object before passing into resolver (for uniformity)

cdimascio:
thank @aviskase 
`router[method.toLowerCase()](path, resolver(basePath, route, this.options.apiSpec));` should be 
` router[method.toLowerCase()](path, resolver(basePath, route, context.apiDoc));`
feel free to create a PR with the change, and i'll get it merged. if not, i'll get it done tonight

aviskase:
yup, will do

cdimascio:
i'll revisit schema. i removed it as part of the `readOnly`/`required` work. 
in order to handle the schema for readOnly vs writeOnly using ajv, i need to manipulate the schema's required property based on the case (readOnly vs writeOnly). i manipulate the required prop once on schema load, rather than per api request for efficiency. though efficient, it has downside, in that two very similar variants of the apiDoc exist. i want to explore other mechanisms before exposing the (not so ideal) implementation detail to the user.

cdimascio:
fixed in v4.3.5. thanks @aviskase !

PR #427, opened 2020-10-27
(fix) #426 custom resolvers are incompatible with options.apiSpec set to path

aviskase:
Also added a simple test for actually checking that resolver works


cdimascio (approved):
Thanks @aviskase!

cdimascio:
@all-contributors add @aviskase for code and test

Issue #320, opened 2020-06-17
Colon in OpenAPI path are mistaken as parameters


yuchenshi:
**Describe the bug**

I'm working on some APIs that follow the format of `/accounts:lookUp` for example. Right now, EOV treats it the same as `/accounts{lookUp}` instead of the literal colon followed by some more words.

BTW, I'm happy to help fixing this.

**To Reproduce**

1. Load an API specification that has `/accounts:lookUp` > `get` as one of the methods. (See example below.)
2. Optionally also set `operationHandlers`.
3. Send an HTTP request to `GET /accounts:bogus`.

**Actual behavior** 

EOV populates the metadata of `/accounts:lookUp` to the bogus request as well, with `pathParams` set to `{lookUp: "bogus"}` (as if `lookUp` were a parameter).

If `operationHandlers` is used, the handler for the lookUp operation is also called for the bogus request. (This is even more problematic if there are multiple operations like `/accounts:lookUp`, `/accounts:signIn` since one can shadow the others depending on ordering.)

**Expected behavior**

As far as I can tell, colons do not have special meanings in OpenAPI v2 / v3 spec, and EOV should just treat it as a literal colon. (This means the implementation should probably escape `:` to `(:)` before passing it to express / pathToRegexp. And then `{pathParam}` replacements can follow.)

**EDIT**: Escaping `:` to `\\:` won't work in Express 4 (which depends on very old `path-to-regexp`), but `(:)` works (which is an unnamed capture that matches only the `:` character).

**Examples and context**

```yaml
paths:
  "/accounts:lookUp":
    get:
       # ...
```

cdimascio:
@yuchenshi thanks for the issue. It will be great to have your help fixing this. Please feel free to investigate. PRs are very welcome.

Also, don't hesitate to reach out to our gitter channel if u need help or want to discuss.

cdimascio:
@yuchenshi, note that is not a valid url character. have you tried url encoding that path segment?

yuchenshi:
@cdimascio I believe `:` is a valid URL character in the path part of an HTTP(S) URL. To quote [RFC3986](https://www.ietf.org/rfc/rfc3986.txt):

```
      path-absolute = "/" [ segment-nz *( "/" segment ) ]
      segment       = *pchar
      segment-nz    = 1*pchar
      pchar         = unreserved / pct-encoded / sub-delims / ":" / "@"
```

[...]

PR #432, opened 2020-10-30
Escape colons in path. Ref #320

siimsoni:
Notes:
* Express.js limits allowed characters in path names to `/[A-Za-z0-9_]/`. This allows splitting by colon. But things may get ugly if path names in the spec are not compatible.
* This does not fix handling of colons in base path of the server. I consider fixing this separately.


cdimascio (approved):
@siimsoni thanks!

cdimascio:
@all-contributors add @siimsoni for code and test

Issue #433, opened 2020-10-30
Colon-separated parameters in base path mapped to invalid express route


siimsoni:
**Describe the bug**

This is similar to #320, except for parameters in base path.

**To Reproduce**

Add server definition in examples to `/test/resources/openapi.yaml`. This will cause 33 tests to fail.

**Actual behavior** 

OpenAPI validator fails when loading the spec.

**Expected behavior**

Colon-separated parameters in base URL will not cause a crash.

**Examples and context**

```
servers:
  - url: /api/{version}:{name}/
    variables:
      version:
        default: v1
        enum:
          - v1
          - v2
      name:
        default: petstore
        enum:
          - petstore
          - storeofpets
```

siimsoni:
On a related note, I didn't find any description about how `express-openapi-validator` should handle the server parameters.
When testing, it appeared I could only use values defined as default in the spec.

cdimascio:
fixed in v4.4.0

PR #437, opened 2020-11-02
feat: handle colons in servers url

cdimascio:
#433


cdimascio:
@siimsoni given your experience with the colons issues, would you mind reviewing this fix. thanks

siimsoni (approved):
This solves the problem! Thanks!

PR #448, opened 2020-11-07
fix: export 406 and rename 413

LEI:
- Add missing external export "NotAcceptable"
- Fix typos in "RequestEntityTooLarge" (breaking change)

For backward compatibility, maybe this line can do the trick, as long as `instanceof` is used to check the error type:

```
  RequestEntityToLarge: RequestEntityTooLarge,
```


cdimascio (approved):
thanks for the fix @LEI.
I'm going to commit as is, despite the breaking change due to the typo.

cdimascio:
@all-contributors add @LEI for code and test

Issue #454, opened 2020-11-11
Custom handling of response validation error


nampas:
**Is your feature request related to a problem? Please describe.**
I'd like to enable response validation in an express app, but I don't want response validation to actually fail the request. In our case, a response validation error likely indicates that we haven't updated our openapi spec correctly. I'd like to log an error, but still return a 200 with the original response.

**Describe the solution you'd like**
It'd be great if there was some option in `validateResponses` to support this, perhaps an `onError` handler that would be invoked if provided instead of the default behavior of throwing.

**Describe alternatives you've considered**
There are probably some shenanigans I could do with re-munging `response.json()`, so that I catch any errors the validator throws. That doesn't seem ideal or straightforward.
Another option is to expose the validators themselves, so that I can call them in my own custom middleware, and handle errors accordingly. This is essentially what we were previously doing with express-openapi-validate, though we've had to abandon that library due to a lack of support for `$ref`s.

cdimascio:
interesting idea. will consider it

cdimascio:
@nampas please feel empowered to implement your idea. i'm happy to review, comment, etc.

Here's a tip to get you started... you will basically need to push an option through to the response validator. then ultimately when the response validation middleware encounters a validation error, it throws  [here](https://github.com/cdimascio/express-openapi-validator/blob/master/src/middlewares/openapi.response.validator.ts#L169). You'll want to customize this behavior. Instead of throwing, invoke a some function that is registered with the `validationResponses` option and return `next()`. 

Some thought is certainly needed around what the function might look like and what a user is expected to implement.
But all in all, this will get you started.

nampas:
Ok thanks @cdimascio . I took a quick stab at it, let me know if you think I'm on the right track.
https://github.com/cdimascio/express-openapi-validator/pull/456

cdimascio:
@nampas closing this out. i believe this is now resolved by your PR #456.

PR #456, opened 2020-11-11
Accept response validation error handler

nampas:
Here's a stab at the feature outlined here: https://github.com/cdimascio/express-openapi-validator/issues/454

This PR adds a new `validateResponses.onError` option. There's no strong contract with the function, but express-openapi-validator will pass the `err` object, as well as the response `data`. Usage might look something like this

```
const handler = (err, data) => {
   console.log('OpenAPI response validation error', err);
};

OpenApiValidator.middleware({
   apiSpec: './spec/openapi.json',
   validateRequests: true,
   validateResponses: { onError: handler }
});
```

### DIFF
--- a/src/middlewares/openapi.response.validator.ts
+++ b/src/middlewares/openapi.response.validator.ts
@@ -13,6 +13,7 @@ import {
   OpenApiRequest,
   OpenApiRequestMetadata,
   InternalServerError,
+  ValidateResponseOpts,
 } from '../framework/types';
 import * as mediaTypeParser from 'media-typer';
 import * as contentTypeParser from 'content-type';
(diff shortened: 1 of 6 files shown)


src/middlewares/openapi.response.validator.ts
```diff
@@ -60,13 +69,23 @@ export class ResponseValidator {
...
+            validators,
+            body,
+            statusCode,
+            path,
+            accepts, // return 406 if not acceptable
+          });
+        } catch (err) {
+          // If a custom error handler was provided, we call that
+          if (err instanceof InternalServerError && this.eovOptions.onError) {
+            this.eovOptions.onError(err, req, res, next, body)
```
nampas, line 83: I think maybe we shouldn't pass `next` here, as the wrapper in `mung` is responsible for invoking that on a successful validation.

cdimascio: can you describe this a bit more

cdimascio: i also wonder if `err` alone is sufficient. is req, res, body needed?

nampas: I think you're right. I don't seen a need for req, res and next. I do think body could be useful though. For example, users may want to extract something from it to log, (e.g. a resource id), or maybe even log the whole thing.

test/response.validation.on.error.spec.ts
```diff
@@ -0,0 +1,76 @@
...
+
+  let onErrorArgs = null;
+  before(async () => {
+    // set up express app
+    app = await createApp(
+      {
+        apiSpec: apiSpecPath,
+        validateResponses: {
+          onError: function() {
+            onErrorArgs = Array.from(arguments);
```
cdimascio, line 20: can you also add a test, where you re-throw the error and/or create a new error. looks like should all work out, but having a test will be good.

cdimascio:
thanks for this @nampas

cdimascio:
one more note, please update the validateResponses section of the README to describe this new option

cdimascio (approved):
@nampas thanks for the feature

PR #459, opened 2020-11-14
Allow ignorePaths to take function or regexp

rgaino:
I have the need of using `ignorePaths` in a way that a RegEx won't suffice, so I added an option to take a function as well as a RegEx. Thanks for your work with the validator!


cdimascio (approved):
@rgaino thanks for this. looks good

Issue #248, opened 2020-02-26
Feature: Extendable fileUploader


Haruki-Mortis:
Hello again :)

I'm going to use multer to upload files, but there are some specific issues with your file uploader.
I want to handle the file upload with multer and then the file should be uploaded to an AWS S3 Bucket.

When I use your fileUploader, the file is only saved in my local folder. It would be really nice if I could add a middleware to your fileUploader so that I can process the new File after multer.

Another option is to set the fileUploader to false and add my own middleware logic. However, if the fileUploader is set false, your validator will no longer validate my openapi definitions for files.

Is there a workaround or am I missing something?

Thank you!

Haruki-Mortis:
* However, if the fileUploader is set false, your validator will no longer validate my openapi definitions for files.

This is wrong. It's more like your validator can't validate the body anymore. It fails if the fileUploader is false and the body property "image" is required.

``` json
    requestBody:
        content:
          multipart/form-data:
            schema:
              description: default
              type: object
              additionalProperties: false
              required:
                -  image
              properties:
                image:
                  type: string
                  format: binary
```

Haruki-Mortis:
@cdimascio sry for the late reply, too.

How about something like this (just a quick example):

## types.ts
```typescript
// extended fileuploader options
export type FileUploaderOptions = {
  multer: multer.Options;
  preMiddleware: FileUploaderMiddlewareOptions[];
  postMiddleware: FileUploaderMiddlewareOptions[];
};

// define a middlware
// optional for specific path(s)
export type FileUploaderMiddlewareOptions = {
  middleware: OpenApiRequestHandler;
  onPath?: RegExp | string;
};

export interface OpenApiValidatorOpts {
  apiSpec: OpenAPIV3.Document | string;
  validateResponses?: boolean | ValidateResponseOpts;
  validateRequests?: boolean | ValidateRequestOpts;
  validateSecurity?: boolean | ValidateSecurityOpts;
  ignorePaths?: RegExp;
  securityHandlers?: SecurityHandlers;
  coerceTypes?: boolean | 'array';
  unknownFormats?: true | string[] | 'ignore';
  fileUploader?: boolean | FileUploaderOptions; // Update this one
  multerOpts?: multer.Options;
  $refParser?: {
    mode: 'bundle' | 'dereference';
  };
  operationHandlers?: false | string | OperationHandlerOptions;
  validateFormats?: false | 'fast' | 'full';
}
```

## openapi.multipart.ts
[...]

PR #464, opened 2020-11-19
248

Haruki-Mortis:
Feature: Extendable fileUploader #248

### DIFF
--- a/src/framework/types.ts
+++ b/src/framework/types.ts
@@ -33,6 +33,8 @@ export type SecurityHandlers = {
 
 export interface MultipartOpts {
   multerOpts: boolean | multer.Options;
+  preMiddleware: OpenApiRequestHandler[];
+  postMiddleware: OpenApiRequestHandler[];
   ajvOpts: ajv.Options;
 }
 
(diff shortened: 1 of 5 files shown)


cdimascio (commented):
@Haruki-Mortis this looks pretty good.
that said, i'm still mulling over whether this is the right way to handle it
can you describe again why we need pre and post middleware?

also, please let me know if you have any thoughts on how this could be handled without requiring pre and post middleware. i understand this may not be possible and (if possible) may require a significant overhaul. that said, now that you've been in the code, you likely have a decent idea of what is and isn't possible. im hoping you might also have thoughts on how file upload might reimagined, so that it can be simpler

thanks a lot for this PR.

setaman:
@Haruki-Mortis @cdimascio I don't want to be bound to `multer.any()` as well and want to be able to provide custom uploader. My approach to implementing this is really simple. We can simply extend the `fileUploader` with an another prop, let's say `uploadHandler`, that is a function returned by `multer().single("image")` for example. Then instead of:
```js
multer.any()(req, res, next) {...}.
```

we can do:

```js
options.multOpts.uploadHandler(req, res, next) {...}
````

This is straightforward to implement and gives the user the ability to use multer without restrictions. I have already implemented and tested this approach exemplary, works fine. What do you say? Is this worth a PR?

cdimascio:
@Haruki-Mortis good idea. this will provide good flexibility

devthejo:
Hello, any news on this ?
I would like to use a custom implementation of multer too, to allow use of upload without multipart, like described here https://swagger.io/docs/specification/describing-request-body/file-upload/
Let me know if a new PR is welcome on this.
PS: Thanks for this great tool you built !

cdimascio:
@devthejo I support the idea and welcome PRs.

Lekky71 (approved):

Issue #396, opened 2020-10-13
Response property casting


Haruki-Mortis:
Hey, I updated from version 3.16.11 to the latest version 4.2.0.
Before the update, the values in the response json were automatically casted to the defined property type.

```typescript
// origin object
{
    createdAt: '1602582518'
}

//object after response.json()
{
    createdAt: 1602582518
}
```

In the openapi spec:
```yml
properties:
    createdAt:
        type: number
        format: int64
        nullable: true
```

This behavior is lost in the latest version.
Is there an option to enable this property casting?

cdimascio:
@Haruki-Mortis yes, this changed and as such, there is not an option to auto cast on response bodies. my feeling is that the option is no longer needed. the validator essentially does was it most appropriate in each situation. i'm happy to be challenged on this if you feel there is a strong need.

with respect to coercion (casting), here is the v4.2.0 behavior
- query, path, header, cookie params are coerced to the declared type. we do this because these inputs can only be provided as strings. as such, coercion is necessary.
- request bodies - types are specified on input e.g.
```jaon
{
  "id": 123,,
   "name": "sam",
   "has_children": true,
}
```

Haruki-Mortis:
@cdimascio thank you for this fast and detailed explanation.
We have a couple of "new" api projects and two of them use something like an automation to generate database queries and filter response bodies.
Your project fits into our projects like a charm and is "the cherry on top".

Unfortunately we can't update your package for these two specific apis. 
The "auto cast" feature is necessary for them.

So I or we would love to see the feature again.

cdimascio:
i'll consider it. the options would likely look something like the following:

```typescript
{
    validateRequests: boolean | ValidateRequestsOpts,
    validateResponse: boolean | ValidateResponsesOpts,
    ...
}

interface ValidateResponseOpts {
       coerceTypes: {
           messageBody: boolean
       }
}

interface ValidateRequestOpts {
       coerceTypes: {
           messageBody: boolean
       }
}
```

```javascript
{
    validateRequests: {
       coerceTypes: {
           messageBody: true, // defaults to false
       }
    },
    validateResponses: {
       coerceTypes: {
           messageBody: true, // defaults to false
       }
    }
}
```

cdimascio:
@Haruki-Mortis is this still something you are looking for? if so, please have a look at PR #451 and provide any feedback that you might have. i believe this will meet your needs.

Haruki-Mortis:
[...]

PR #468, opened 2020-11-25
Allowing request body to be coerced

MadMango:
We talked about implementing it in #387 and #396

Hope the code is fine, let me know if it needs any tweaks, it's still false as default so it shouldn't affect current behaviour in anyone's projects.

Added tests to cover new features.

I've also made a change to the global `coerceTypes` setting, it wasn't affecting anything apart from setting it to true on response validation which seems counterintuitive, especially that now we have both response and request validation so I just removed the code that seemed redundant on [these lines](https://github.com/cdimascio/express-openapi-validator/pull/468/files#diff-45e81b10790fecfe1b0108e22869e679c99c58f5a51d8d2b35c5442f37a5cb60L320-L330)

### DIFF
--- a/README.md
+++ b/README.md
@@ -543,7 +543,7 @@ Determines whether the validator should validate requests.
   }
   ```
 
-    `allowUnknownQueryParameters` is set for the entire validator. It can be overwritten per-operation using
+  `allowUnknownQueryParameters` is set for the entire validator. It can be overwritten per-operation using
   a custom property `x-allow-unknown-query-parameters`.
 
   For example to allow unknown query parameters on ONLY a single endpoint:
@@ -561,6 +561,7 @@ Determines whether the validator should validate requests.
         responses:
           200:
             description: success
+    ```
 
   **coerceTypes:**
 
--- a/README.md
+++ b/README.md
@@ -546,7 +546,7 @@ Determines whether the validator should validate requests.
   **coerceTypes:**
 
   - `true` - coerce scalar data types.
-  - `false` - (**default**) do not coerce types. (almost always the desired behavior)
+  - `false` - (**default**) do not coerce types. (more strict, safer)
   - `"array"` - in addition to coercions between scalar types, coerce scalar data to an array with one element and vice versa (as required by the schema).
   For example:
 
--- a/README.md
+++ b/README.md
@@ -543,6 +543,25 @@ Determines whether the validator should validate requests.
   }
   ```
 
+    `allowUnknownQueryParameters` is set for the entire validator. It can be overwritten per-operation using
+  a custom property `x-allow-unknown-query-parameters`.
+
+  For example to allow unknown query parameters on ONLY a single endpoint:
+
+  ```yaml
+  paths:
+    /allow_unknown:
+      get:
+        x-allow-unknown-query-parameters: true
+        parameters:
+          - name: value
+            in: query
+            schema:
+              type: string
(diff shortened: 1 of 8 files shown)


README.md
```diff
@@ -543,6 +543,19 @@ Determines whether the validator should validate requests.
...
+  **coerceTypes:**
+
+  - `true` - coerce scalar data types.
+  - `false` - (**default**) do not coerce types. (more strict, safer)
+  - `"array"` - in addition to coercions between scalar types, coerce scalar data to an array with one element and vice versa (as required by the schema).
+  For example:
+
+  ```javascript
+  validateRequests: {
+    allowUnknownQueryParameters: true;
```
cdimascio, line 555: update example to demonstrate coerceTypes rather than unknown query params

src/framework/types.ts
```diff
@@ -42,6 +42,7 @@ export interface RequestValidatorOptions
 
 export type ValidateRequestOpts = {
   allowUnknownQueryParameters?: boolean;
+  coerceTypes?: boolean | 'array';
```
cdimascio, line 45: There are two cases for request coercion, i.e. query parameters and body
response coercion has only one type, body
in order to satisfy both request coercion cases, there could be two properties under coerceTypes. this would give complete control, but it results in a lot of options (which i don't like).

all that said, i wonder if it reasonable to go with what you have...
i don't ever see a reason why one wouldn't coerce query params. its essentially required since they all come in as strings. that said, there are some edge cases e.g. a complex json object passed as query param value. in such cases, coercion applies. thinking on this, i think it makes sense to apply the same coercion rules. if so, this change works

thoughts?

MadMango: Are you saying this setting controls both query params and body? I only meant to allow body to be coerced.

README.md
MadMango: fixed, thanks

src/framework/types.ts
MadMango: I just had a look and it doesn't look like that to me, I may be wrong though! I think it only uses that option to validate body: https://github.com/MadMango/express-openapi-validator/blob/master/src/middlewares/openapi.request.validator.ts#L49

cdimascio: Ya. I was wondering wether this should be an option for query params, path params, headers, cookies
Ultimately, it should not as those inputs can only come in as strings hence, they must be coerced. 

All in all, I posed the question becasuse query params, path params, headers, and cookies will behave differently than bodies. I imagine this may be confusing to folks. 

That says, I believe this is the correct behavior and just needs documenting in the README

thoughts?

MadMango: Yeah I think all of those should be coerced and I believe they are at the moment and we're only leaving the option for coercing body. Should I make it very clear in the readme that his option only applied to the request body?

cdimascio: Ya. let's add that to the readme. i'll get this merged in soon.
Thanks for your work, @MadMango

MadMango: I've updated the README, please have a look and see if you're happy with the wording, tried to keep it consistent

cdimascio: look great!

cdimascio (approved):

cdimascio:
@ex1st this last commit didn't make it into the merge, can you put up a new PR. we can review it via a new flow. thanks!

PR #477, opened 2020-12-01
feat: Specify allow unknown params on a per-operation level

JacobLey:
I came across a use case where I should allow unknown query params (An OAuth RedirectURI which may be given extra unnecessary context, but since "consumers" of this API are not calling it directly, I don't want to reject the entire request for providing extra query params).

However as a general rule I like to go with the default `allowUnknownQueryParameters=false`. Since that is a validator-wide option, it wasn't possible to break the rules for just one operation.

I took a stab at implementing a per-operation override using OpenAPI's custom properties. Now I can keep strict rules on the rest of my endpoints.

I included tests and documentation as well. I figure the naming is pretty obvious, and low-risk of colliding with existing use cases that don't intend to do the exact same thing.

I can create an actual issue to discuss other alternatives if this is not an acceptable solution.

### DIFF
--- a/src/framework/types.ts
+++ b/src/framework/types.ts
@@ -161,7 +161,6 @@ export namespace OpenAPIV3 {
     deprecated?: boolean;
     security?: SecurityRequirementObject[];
     servers?: ServerObject[];
-    'x-allow-unknown-query-parameters'?: boolean;
   }
 
   export interface ExternalDocumentationObject {
--- a/src/framework/types.ts
+++ b/src/framework/types.ts
@@ -161,6 +161,7 @@ export namespace OpenAPIV3 {
     deprecated?: boolean;
     security?: SecurityRequirementObject[];
     servers?: ServerObject[];
+    'x-allow-unknown-query-parameters'?: boolean;
   }
 
   export interface ExternalDocumentationObject {
--- a/src/middlewares/openapi.request.validator.ts
+++ b/src/middlewares/openapi.request.validator.ts
@@ -189,13 +189,13 @@ export class RequestValidator {
     if (discriminator) {
       const { options, property, validators } = discriminator;
       const discriminatorValue = req.body[property]; // TODO may not alwasy be in this position
-      if (options.find((o) => o.option === discriminatorValue)) {
+      if (options.find(o => o.option === discriminatorValue)) {
         return validators[discriminatorValue];
       } else {
         throw new BadRequest({
           path: req.path,
           message: `'${property}' should be equal to one of the allowed values: ${options
-            .map((o) => o.option)
+            .map(o => o.option)
             .join(', ')}.`,
         });
       }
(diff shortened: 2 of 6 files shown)


src/middlewares/openapi.request.validator.ts
```diff
@@ -208,14 +213,11 @@ export class RequestValidator {
       keys.push(key);
     }
     const knownQueryParams = new Set(keys);
-    whiteList.forEach((item) => knownQueryParams.add(item));
+    whiteList.forEach(item => knownQueryParams.add(item));
     const queryParams = Object.keys(query);
     const allowedEmpty = schema.allowEmptyValue;
     for (const q of queryParams) {
-      if (
-        !this.requestOpts.allowUnknownQueryParameters &&
```
JacobLey, line 216: This method `processQueryParam` is only called in one place, and the `allowUnknownQueryParameters` boolean is checked there beforehand, so this check is redundant. (Also with my change, it doesn't check the correct scope)

src/framework/types.ts
```diff
@@ -161,6 +161,7 @@ export namespace OpenAPIV3 {
     deprecated?: boolean;
     security?: SecurityRequirementObject[];
     servers?: ServerObject[];
+    'x-allow-unknown-query-parameters'?: boolean;
```
cdimascio, line 164: For now, I'd prefer not to add the vendor extension onto the OperationObject. Can you make this work without it (should be doable with a little TypeScript pain)

src/middlewares/openapi.request.validator.ts
```diff
@@ -125,7 +130,7 @@ export class RequestValidator {
 
       mutator.modifyRequest(req);
 
-      if (!this.requestOpts.allowUnknownQueryParameters) {
+      if (disallowUnknownQueryParameters) {
```
cdimascio, line 133: !allow...

src/middlewares/openapi.request.validator.ts
cdimascio: good catch! thanks for cleaning it up

cdimascio (commented):
nice idea. thanks for implementing the feature.
have a few minor comments. it will also be great if you can update the README to include this

src/framework/types.ts
JacobLey: Removed.
Using `!!` Typescript automatically converts the `any` to `boolean`. Which seems like enough to me.

I could explicitly cast to boolean, or perhaps even better check `typeof === 'boolean'` but that seems like a bit overkill.
I guess the caveat is `x-allow-unknown-query-parameters: definitely not a bool` will get converted to `true`

JacobLey:
> it will also be great if you can update the README to include this

I updated the existing `validateRequests->allowUnknownQueryParameters` section of the README. I can add another section (or expand on that section) as well if that isn't enough

cdimascio:
> > it will also be great if you can update the README to include this
> 
> I updated the existing `validateRequests->allowUnknownQueryParameters` section of the README. I can add another section (or expand on that section) as well if that isn't enough

Yep. i somehow overlooked it. Thanks!

cdimascio (approved):
Awesome!

cdimascio:
@JacobLey this is available in v4.8.0

PR #490, opened 2020-12-15
Fix typos in readme

Parnswir:
Fixes three typos and updates the position of an example in the FAQs that seems to have been misplaced when a new FAQ item was added.


cdimascio (approved):
@Parnswir thank you!

Issue #353, opened 2020-08-20
Serialize in request validation and authorize object when toJSON() matches to the format


pilerou:
**Is your feature request related to a problem? Please describe.**
I would like to do 2 things :
- serialize String data to Object when I receive it from request
- deserialize object when i want to return the response and validate my response

I have two kinds of objects in my mind :
- Date objects
- Mongodb ObjectID

**Describe the solution you'd like**
It could be great to allow developers to add some adapters that could be considered by express-openapi-validator on request and on response.
For example, i could declare in my openapi components section :
```yaml
  components: 
    schemas: 
      ObjectId:
        type: "string"
        pattern: "^[0-9a-fA-F]{24}$"
      Date:
        type: "string"
        format: "date"
      DateTime:
        type: "string"
        format: "date-time"
```

then i could configure my OpenAPIValidator with those adapters :
```javascript
OpenApiValidator({
  apiSpec: './test/resources/openapi.yaml',
  validateRequests: true, // (default)
  validateResponses: true, // false by default
  formatAdapters : {
     '#/components/schemas/ObjectId' : {
        serialize : o => new ObjectID(o)
        deserialize : o => o.toJSON()
    },
    '#/components/schemas/Date' : {
        serialize : o => new Date(o)
        deserialize : o => o.toISOString().slice(0,10);
    },
    '#/components/schemas/DateTime' : {
        serialize : o => new Date(o)
        deserialize : o => o.toISOString();
    } 
  }
})

```

**Describe alternatives you've considered**
I already tried to hack the formats.js by adding new formats and a "validate" description
I also used in the past (2014-2015), a library, "tdegrunt/jsonschema" ,in which I added a hook to do this.
I manage to do this but I need to parse twice the JSON body : once with jsonschema then once with express-openapi-validator.

**Additional context**
I saw some issues related to this issue even if they are not explained the same way. I think this solution would be adaptable : 
- https://github.com/cdimascio/express-openapi-validator/issues/246

pilerou:
Actually I didn't manage to make it work.
- I can hack deserialization with `JSON.parse(JSON.stringify(object)) `.
It's good for converting `Date` object to `date-time` json schema. It won't solve the case of a `date` json schema
- I didn't manage to serialize json to objects. I don't know where to do it in the module and I think it's related to AJV usage. I don't know very much how to change something in it.

If you give me some clues, I'm OK to try to implement this feature.

pilerou:
[...]

Issue #465, opened 2020-11-23
How do you handle mongodb ObjectId ?


balazssoltesz:
Hello,

im trying to use express-openapi-validator with mongodb, but I can't figure out how should I handle ObjectId fields.
I have to specify id fields in the openapi.json as string, and in res.json() it will be converted to string as well, but if i set validateResponses to true then it will throw an error for me (id should be string).
Any idea how to handle properly mongodb _id fields?

cdimascio:
You will want to send _id as a string. If you are returning the objectid as is, it's likely to be object and not a string, hence why the validator complains that it should be a string.

balazssoltesz:
Yes I understand but practically it will be sent as string cause res.json() convert it to string. Should is iterate over the whole collection to convert any ObjectId field to string manually to avoid validator error?

pilerou:
I managed objectid as described here
https://github.com/cdimascio/express-openapi-validator/issues/353

pilerou:
@cdimascio it could be great to add it. If you want I could try to push for a merge request

cdimascio:
hey all @pilerou has propsed a solution https://github.com/cdimascio/express-openapi-validator/pull/493
please have a look. feedback is greatly appreciated

cdimascio:
@balazssoltesz this is now resolved by using the serDes option. See serDes doc. 

@pilerou has also provided a serializer/deserializer lib that can be used with the validator to handle mongodb object.

See the readme for details on both

Issue #288, opened 2020-05-06
Response validation: Automatically coerce date data type to ISO string


xeoneux:
**Is your feature request related to a problem? Please describe.**
When validating responses, if the data type is a js `Date` object, the validation of type string with format date-time fails.

```js
{
  updatedAt: new Date() // fails
}

{
  updatedAt: "2010-13-30T23:12:35Z" // passes
}
```

**Describe the solution you'd like**
The lib should try to coerce the type to ISO string and then try validation on it.

```js
{
  updatedAt: new Date() // should pass
}
```

**Describe alternatives you've considered**
I have to programmatically search through all the object to find a date data type and call the `toISOString` method on it to pass the output validation.

**Additional context**
None

cdimascio:
@xeoneux is this occurring with response validation?

xeoneux:
This is with response validation, didn't check the request part.

cdimascio:
@xeoneux I did some investigation here. This is a bit tricky to solve. I'm not sure a satisfactory solution exists within Ajv.

I tried to create a custom Ajv keyword. A keyword will enable me to manipulate the `Date` object and convert it to a string e.g. `toISOString()`, The problem is described [here](https://github.com/epoberezkin/ajv/issues/141#issuecomment-270692820) and [here](https://github.com/epoberezkin/ajv/issues/192) note:  
> The only caveat at the moment is that custom keywords will be validated AFTER standard keywords in the same schema object, ...

AFTER is the issue. Ultimately, validation will fail despite modifying the data because an `object` is being passed to a schema that expects a `string` and that validation will fail prior to modifying the data. There is no current way to modify the data prior to validation. 

I certainly agree that using `new Date()` in this context is convenient, but there does not seem to be an obvious solution, other than assigning e.g. `updateAt` to a string e.g. `new Date().toISOString()`

The following is the keyword I was experimenting with
```
    // if this keyword is set, convert a javascript Date to its iso string
    // TODO validate the date, use Date.parse
    ajv.addKeyword('x-eov-date-as-iso-string', {
      modifying: true,
      compile: function (sch, parentSchema) {
        const match = (<any>parentSchema).format === 'date-time' && sch;
        return match
          ? function(value, objectKey, object, key) {
              object[key] = new Date(value).toISOString();
              return true;
            }
          : function() => return true;
      },
    });
```

cdimascio:
keeping this open, perhaps folks have alternative ideas

[...]

Issue #246, opened 2020-02-25
Date is not validated correctly as type: string, format: date-time


severinwullschleger:
Hi

I want to return the _created_ and _lastModified_ timestamp from the database which I define in my ORM with type _Date_.

If I print the response object in my express router before returning it, it looks like this:
```
User {
   id: '1fe87c1d-5d31-4426-8757-aa5e62d2344e',
   lastModified: 2020-02-21T09:26:43.603Z,
   created: 2020-02-21T09:25:48.151Z,
```

If I call the endpoint I get the following reponse object:
```
{
  "data": [
    {
      "id": "1fe87c1d-5d31-4426-8757-aa5e62d2344e",
      "lastModified": "2020-02-21T09:26:43.603Z",
      "created": "2020-02-21T09:25:48.151Z"
    },
...
```
-> correctly stringified

As soon as I set the validateResponses option in the OpenApiValidator on true, I get the following error message:
```
{
  "error": ".response.data[0].lastModified should be string, .response.data[0].created should be string,
```
... although it is a string in the upper example

martinbark:
I just hit this same issue today.  It's because the date is an object not a string.  At the point the response it validated it's still a Date object.  When you print it [date.toJSON()](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date/toJSON) is called to convert it to a string.  It's not specific to Date, you would see this issue on any object that has a [toJSON()](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/JSON/stringify#toJSON_behavior) function

The only reliable way i found to work around this is to `JSON.stringify()` then `JSON.parse()` the response body before it's validated.  This means the response body is now formatted as the destination will see the message before it's validated.

I added this code after OpenApiValidator was been installed as a workaround
```
app.use((req, res, next) => {
  const original = res.json;
  res.json = function jsonp(_json) {
    return original.call(this, JSON.parse(JSON.stringify(json)));
  };
  next();
});
```

severinwullschleger:
Thanks a lot. The workaround worked for me too.

electrotype:
Using `send` seems to be required for me instead:

```
  app.use((req, res, next) => {
    const original = res.send;
    res.send = function jsonp(json) {
      return original.call(this, JSON.parse(JSON.stringify(json)));
    };
    next();
  });
```

Dakuan:
horrible

[...]

PR #493, opened 2020-12-18
coerceComponents feature for middleware in order to cast data

pilerou:
This pull request adds an attribute to middleware configuration : `coerceComponents`.
It allows to cast : 
- some string to objects (Date, MongoDB ObjectId) on request 
- objects to string (or other type) on response

I added an example project and a suggestion for README.MD


pilerou:
This pull request could close #353 #465 #288 #246

electrotype:
Sadly, doesn't seem to work for me for a `date-time` field...

In your tests, please add a case where you validate a response consisting of an object with a property of type `Date` and it passes the validation as a `date-time` field in the Open API spec.

cdimascio:
@pilerou it will be great to add test cases that at least cover the three examples you highlight in the example packages

here is an example [test file](https://github.com/cdimascio/express-openapi-validator/blob/master/test/path.params.spec.ts) to help get your started

pilerou:
Sorry for the first version. It still had a bug. I just fixed it, I modified the example to send a date.
I also added a test file for this feature that send a response with : 
- an ObjectID id, 
- a creationDate in yyyy-mm-dd 
- a creationDateTime in ISO date format

electrotype:
Thank you so much for this PR @pilerou . Merci beaucoup!

I think I understand how your code works. You can tell `ajv` to run a specific code _for a given schema_, is that correct?

I've been able to make your code work in my project, but with a change on my side.

In my spec file, this _doesn't_ work:

```
components:
  schemas:
    User:
      properties:
        creationDate:
          type: string
          format: date-time
```

But this _does_ work:

```
components:
  schemas:
    DateTime:
      type: string
      format: date-time
    User:
      properties:
        creationDate:
          $ref: '#/components/schemas/DateTime'
```

In other words, you need to define a schema for the components you want to tweak, is that correct?

Do you see any way to make your new feature work with my first exemple too? Don't get me wrong, I think your PR should be merged now, and I will definitively use its `coerceComponents` feature! But being able to validate an inline `string/date-time` type would be "la crème de la crème"!

pilerou:
I agree with you about the perfect solution that would be to use the `format` or `pattern` attribute.
Unfortunatly, AJV doesn't allow to coerce on this field.
I didn't manage to override `format` keyword.

At the end, I convinced myself that it's finally a good practice to have a generic component type definition for Date, Date-time or ObjectId... :)

The feature can solve many issues but I do agree that it's not very intuitive for beginner on this module. I didn't find a better way to do it easier. :S

electrotype:
I think this is an acceptable solution (maybe even the only one)! I would suggest that there is a small section in the documentation to explain it, though.

In issue #288 I said I would give a bounty for a fix so, as soon as the PR is merged, let me know what you want me to do with the money. Send me your Paypal address if you want. Thanks again!

pilerou:
I didn't see for the bounty purposal. 
It's a pleasure if someone can use this feature. 
Thanks for the money but I prefer to keep the opportunity to ask your help on this module when I'll need to :)

electrotype:
The money will go to the project itself then ("Sponsor" link). Cheers!

cdimascio:
@electrotype thank you :) very generous. 

once it's merged in, i will plan to reroute the full bounty to @pilerou. he did the work and deserves it.

@pilerou DM me on [twitter](https://twitter.com/carminedimasio) or [linkedin](https://www.linkedin.com/in/carmine/), so we can sort out how to get it to you.

cdimascio:
@pilerou, we have serialize/deserialize backwards

`serialize` should convert the in-memory object e.g. `Date` to a `string`

`deserialize` should take a `string` and convert it to `Date`

cdimascio:
review notes:
- rename `coerceComponents` to `schemaObjectMapper`
- serialize should have form object to string 
- deserialize should have form string to object

pilerou:
Thanks for the review. I change it and I push.

pilerou:
I did the changes :
- Changes `coerceComponents` to `schemaObjectMapper`
- Fix function names serialize and deserialize => To be clear for everybody on which functions are executed on requests and responses, it's now `deserializeRequestComponent` and `serializeResponseComponent`
=> Unit tests seem OK and I also changed documentation, files name... I tested at home the example. It's also OK "on my computer"

cdimascio:
@pilerou i made modifed the types to use `serialize` `deserialize` and updated the README to be more description about the usage. let me know if it seems clear or if i should add more detail to ensure folks know when to use serialize vs deserialize

cdimascio:
while looking making these modifications, another question arose.

what is the behavior if a user sets `validateResponses` to false, but defines `schemaObjectMapper`?

pilerou:
@cdimascio All modifications are good to me. Your english is also much better than mine :+1: 
I tried with validateResponses to false.
As I thought, It's not working. `serialize` function is called on response schema validation process. If there's no schema validation on response, there's no serialization.

I see 2 options :
- When `schemaObjectMapper` is set and `validateResponses` or `validateRequests` aren't, we force those attributes to `true` at `middleware` creation
- We add a comment explaining `validateResponses` must be set to true.

I prefer the first solution. What do you prefer ?

cdimascio:
@pilerou gotcha. in that case, it's best to just throw an error right away and let the user sort out how to proceed.

The validation checks can be added [here](https://github.com/cdimascio/express-openapi-validator/blob/a542ece187b6521e865e1e28a287216b2bade998/src/openapi.validator.ts#L311).

All in all, if either `validateRequests` or `validateResponses` is not enabled, throw an `Error` with a message like "schemaObjectMapper requires that validateRequests and validateResponses options are both enabled."

Also, do you see any way that we could provide this capability without requring `validateResponses`?

There are many users who enable `validateResponses` only in their dev environments, then disable it in prod. This is different from `validateRequests`, where users always have it enabled in all environments.

pilerou:
I understand.
I think about it and I try to find a solution for validateReponses. No time this week to work on it but I hope for the end of 2020 😁

cdimascio:
Thanks. I'll think on it as well.

cdimascio:
@all-contributors add @pilerou for code, test, and ideas

cdimascio:
@pilerou @electrotype 
i was thinking about this a bit more. the issues only arises when `validateResponses` is enabled. The reaseon for this is because repsonse validation occurs on the in-memory representation of the response object prior to that object being sent to the client via e.g. `res.json`. 

Thus, when the response validation runs, the `Date` object is still a `Date` object, thus the validator sees an object, yet expects a string, then returns a validation error. This is the core of the problem. 

On the other hand, when `validateResponses` is disabled, the schema is never validate, then when `res.json` is called on the response object, the JSON serialization kicks in and `Date` object becomes a string.

An easy way to solve this, is to serialize the response object prior to validation e.g.

```javascript
    body = JSON.parse(JSON.stringify(body));
    const valid = validator({
      response: body,
    });
```    
This solves the problem. Though simple, it requires serialization of response object twice. Once before validation, and again when `res.json` is called.

Here is the code: https://github.com/cdimascio/express-openapi-validator/pull/498. This doens't require mappers but has the side effect i noted above

It would be great to handle this with something like the above, but somehow avoid the double serialization. It's simpler, doesn't complicate the API/options, and behaves the same as it would if

electrotype:
> body = JSON.parse(JSON.stringify(body));
> This solves the problem. However, when `validateResponses` is enabled, it requires serialization of response object twice. Once before validation, then when `res.json` is invoked.

@cdimascio This is _exactly_ what we are trying to avoid! :-) See https://github.com/cdimascio/express-openapi-validator/issues/288#issuecomment-738060730 . Yes it works, but you then se/deserialize _every response_, with or without dates, simply because it is required for this validation to work.

The other "workaround" is for the appliaction code itself to know that it has to convert the date properties before sending the responses. It is faster because it knows the entities so can convert specific properties without se/derializing everyting. I really do not like this solution because the application would do this simply because `express-openapi-validator` is used. This is way too tightly coupled!

From what I understand of the code, the _real_ solution would be the make a PR to `ajd` for it to not validate the type (string/object) of a property so early and in a so hardcoded way.

If you are able to hack `ajd` to prevent that type validation to be done so early, that would be impressive and is clearly a good challenge.

cdimascio:
@pilerou i refactored the schema preprocessor to traverse the request and response schema in one pass. it effectively traverses one schema, but also updates to both along the way.

This improvement makes it possible to associate the custom keyword with any node in the spec (rather than those we register via property like `schemaObjectMapper`. By hooking into the preprocessor, we can for example match nodes like the following:

```yaml
type: string
format: date-time
```

This is essentially what @electrotype was expecting. It's certainly more idiomatic. I'd like to take this route.

PR https://github.com/cdimascio/express-openapi-validator/pull/499 implements this improved preprocessor.
(Note, the PR is large, but all you need to be concerned with is one method, [`handleSerDes`](https://github.com/cdimascio/express-openapi-validator/blob/2bbe0d1841b2fc34efec3cd4086d52caf3b648cf/src/middlewares/parsers/schema.preprocessor.ts#L321-L333).

I'm hoping that you might fork the PR branch and integrate something similar to what you've done. One difference being that the user can define the types as usual (just specify `type` and `format`.

Here's what I'm thinking:

For formats defined in the OpenAPI spec (`date-time`, `full-date`), do the following:
1. Implement the keyword in  a manner similar to what you've done in `ajv/index.ts`
_Note: this [code](https://github.com/cdimascio/express-openapi-validator/blob/2bbe0d1841b2fc34efec3cd4086d52caf3b648cf/src/framework/ajv/index.ts#L117-L131) is not needed_
2. In  [`handleSerDes`](https://github.com/cdimascio/express-openapi-validator/blob/2bbe0d1841b2fc34efec3cd4086d52caf3b648cf/src/middlewares/parsers/schema.preprocessor.ts#L321-L333), check if e.g. `type: string format: date-time` is present, if so, assign the keyword + function, 
_Note: `state.kind` will have a value `req` or `res` which tells u if you are modifying the request or response's api doc_
3. If we can do this with a `serializer` only. (see my note at the bottom of this comment) then Implment the `serializer` for `Date`. It should be automatically used anytime `type: string format: date-time` is encounted. We could also create our own `EovDate` that implements the appropriate `toJSON` to ensure the proper stringified date is serialized by express.

If we require a `serializer` and `deserializer` then, we'll require the user to provide them via an option like `schemaObjectMapper`


For formats not defined in the OpenAPI spec e.g. MongoDb `ObjectId`
1. Use something like `schemaObjectMapper` to enable users to provide custom functions. Basically, similar to what you've done. Note: if we can do this with a `serializer` only (see my note at the bottom of this comment), then we don't need the user to provide a deserializer.
2. n  [`handleSerDes`](https://github.com/cdimascio/express-openapi-validator/blob/2bbe0d1841b2fc34efec3cd4086d52caf3b648cf/src/middlewares/parsers/schema.preprocessor.ts#L321-L333), check if e.g. `type: string format: my-mongo-objectid` is present, if so, assign the keyword + function.

A user might specify a custom mongo object id custom format like so e.g.
_Note: this is nice as its idiomatic_

```yaml
type: string
format: mongo-objectid
```

and we can trigger the custom ser/des


As i noted above, we may be able to do this using _**only**_ a `serializer` (no deserializer). AJV validaties JSON-schema. In AJV, I believe we can validate a `Date` of type `object`. For example, if we get `Date` object today. It fails validation because the schema states `type: string`. With the new preprocessor, we could change `type: string` to `type: object`. In theory, this would validate the `Date` object, assuming it passes, it would serialize using its `toJSON()` method (just as it does without validation). The nice thing here, is that we get the same serialization whether or not we use`validateReponses`. To implement different `Date` formats (rather than what `Date.toJSON()` returns, we could have the keyword function store the `Date` in custom `Object` and attach `toJSON()` and `toString()` method that serializes the `Date` using the appropriate representation (defined by `format`

cdimascio:
Good news! we can definitely do this with a single serializer. Example [here](https://github.com/cdimascio/express-openapi-validator/blob/80d9895295b38c2946adb3ad95326a919ce62644/src/middlewares/parsers/schema.preprocessor.ts#L47-L52) in the new PR

electrotype:
Feedback left on https://github.com/cdimascio/express-openapi-validator/pull/499
Thanks to both of you, again.

cdimascio:
@electrotype have a look at https://github.com/cdimascio/express-openapi-validator/pull/499
I'm thinking this may be the final form. Note that the new behavior is opt-in. This will ensure existing scenarios continue to work as is.

I'm curious to get your thoughts for the next major version. Should this be opt-in? or default behavior.

Note that its essentially either or. if the serializer is enabled, then you **MUST** specify `Date` objects in your responses if the serializer is enable for `date` and/or `date-time`. if the serializer is not enabled, then you **MUST** specify a date as a `string`. One could enable the serializers, then specify date `strings` as long as `format` is not specified. In any case, interested to get your take

electrotype:
Happy Holidays to you too.

First, excuse me because I know I'm not as deep into the code itself as you (both) are. I currently do not have the time to invest in understanding everything involved in this issue, as much as I would like. Sorry about that, I may try to contribute more to this project one day.

I understand your question and why you are not sure about the best way to go. If I enable the `Date` serializers, then I could _not_ send a Date as a string anymore:

```
res.send({
  created_at: '2020-12-25T03:36:34.706Z';
});
```

And I agree this is not the ideal. It *is* confusing. I think maybe the root issue of all that is that the validation is not made on what is actually sent to the client (a Date will always result in a string at the end!), but on the raw response your controllers (or whatever) are sending in the first place...

My random ideas/notes:

1) Would it be possible to make the validation run on the serialized response instead of on the raw response? So later in the process on sending a response? I guess this would required an extra deserialization though.

2) Instead of converting the type of a Date from `string` to `object` as the main trick for the validation to work, couln't you, in a way or another, use some kind of [oneOf](https://swagger.io/docs/specification/data-models/oneof-anyof-allof-not/) (also:  https://swagger.io/docs/specification/data-models/data-types/#mixed-type ), in order to specify that a raw Date property can be both a string _or_ an object?

3) In addition to converting the type of a Date from `string` to `object` as the main trick for the validation to work, could you also _create a new schema on the fly_ and use this schema to allow both `string` and `object`?

4) Question: Does the original solution (from @pilerou) work with strings too? If I create a `DateTime` schema instead of defining an inline `format: date-time`, could I send Dates as strings? I didn't check.

Dates really are a tricky type in Javascript/JSON!

cdimascio:
1. is difficult to get correct, particularly given response are streamed.
2. yes! we can use the mixed variant that OpenAPI notes as `incorrect`. It's validate JSON schema thus will work
3. 2 above will do it
4. no, same problem (but it could work with 2 above)

all in all, we can do 2. are remove the `serializer` option. Note that option will likely return if @pilerou adds support for custom `serializer`s, so that folks can handle e.g. Mongo `ObjectId`s and more

cdimascio:
@all-contributors add @electrotype for ideas

electrotype:
Can't wait to see the final PR! You guys are doing an incredible job... Listening to your users and such. Cheers!

And don't worry about the bounty and where it will go, just make sure the solution is the best one. I will make sure everybody involved here has my official "thank you very, very much" appreciation... The hard part is sometimes simply to make things move. :-)

cdimascio:
`v4.10.0` is live and provides a solution for `date-time` and `date` that does not require any user configuration.

@pilerou it will be fantastic if you can add the functionality for `ObjectId` once you're time frees up again.

A path toward a solution might look something like this:

1. Create a new option `validateResponses.serializers` It's shape will look something like `schemaObjectMapper` (perhaps a list?) e.g.

    ```javascript
    validateResponses: {
      serializers: [{
         format: 'mongodb-objectid',
         serializer: (o) => o.toString(),
      }]
    }
    ```

2. Modify [handleSerDes](https://github.com/cdimascio/express-openapi-validator/blob/master/src/middlewares/parsers/schema.preprocessor.ts#L334-L353) to handle custom serializers for a registered format whose input is a JavaScript object, not a string e.g. MongoDb `ObjectId`

electrotype:
I'll be trying 4.10.0 very soon. Thank you!

pilerou:
Hi.
Things are moving forward. That's great.
I'm looking to changes you made.
I'm not sure the new mecanism allows the same. Could you confirm ?

**Questions**
- Does it also `deserialize` fields on request ? When I get a `string` in request with `date-time` `format`, does it cast the string to `Date` object ?
- For `ObjectId` components, I don't describe it with `format` description field. I use `pattern` field. New PR code use `format` field only. We could add other field (such as `pattern`) each time we need to but it will get heavier and heavier.

**Implementation examples : **
- For ObjectId : 
  - I'll need to add an `unknownFormats` for `mongodb-objectid` ? 
  - Then I'll need to add a new option to `validateResponses` as described in your solution ? 
  - Do I need to configure something on request validation ? 
- I call a request with objectId in url : /user/5fe65b0dc4a82ae0cb4d914a 
  ==> With no `deserialize` function, I'll have a string in my route function or I'll have (and don't forget) to cast it to ObjectId in my route function. That's what I wanted to avoid.

**Suggestion**
 - I think we should configure Serialize/deserialize as a Map instead of an Array. We can only add on serialize function for each `format`. With the "Array" conception, it would allow users to add many configuration for a `format`. I began to make some changes that make it more generic. The module would configure basic serializers for `date` and `date-time` and we could add new serializers in ValidateResponseOpts. Thos serializers would be added or could override basic serializers.

Here how I changed it in `schema.preprocessor.ts` (I also changed `types.ts` to make it compile) :

```javascript
export class SchemaPreprocessor {
  private ajv: Ajv;
  private apiDoc: OpenAPIV3.Document;
  private apiDocRes: OpenAPIV3.Document;
  private responseOpts: ValidateResponseOpts;
  private serializers: { [p: string]: Serializer; };
  constructor(
    apiDoc: OpenAPIV3.Document,
    ajvOptions: ajv.Options,
    validateResponsesOpts: ValidateResponseOpts,
  ) {
    this.ajv = createRequestAjv(apiDoc, ajvOptions);
    this.apiDoc = apiDoc;
    this.responseOpts = validateResponsesOpts;
    if(this.responseOpts && this.responseOpts.serializers)
    this.serializers = {
      ...basicSerializer,
      ...this.responseOpts.serializers
    }
  }
```

```javascript
const basicSerializer = {
  'date-time': {
    serialize: (d: Date) => {
      return d && d.toISOString();
    },
  },
  'date': {
    format: 'date',
    serialize: (d: Date) => {
      return d && d.toISOString().split('T')[0];
    },
  }
};
```

```javascript
private handleSerDes(
    parent: SchemaObject,
    schema: SchemaObject,
    state: TraversalState,
  ) {
    if (state.kind === 'res') {
      if (schema.type === 'string' && !!schema.format && this.serializers[schema.format]) {
        schema['x-eov-serializer'] = this.serializers[schema.format];
      }
    }
  }
```

cdimascio:
@pilerou 
> Does it also deserialize fields on request ? When I get a string in request with date-time format, does it cast the string to Date object ?

The new feature does not affect the request at all, thus
- it does not require a user option (and remains backward compatible)
- it does not attempt to deserialize any value in a request
- it also does not block the addition of deserializers. this is still possible.

> For ObjectId components, I don't describe it with format description field. I use pattern field. New PR code use format field only. We could add other field (such as pattern) each time we need to but it will get heavier and heavier.

- The current code only checks `format`, however, `pattern` could also be checked.  Note that the new implementation did **_not_** add or change any user-facing options or APIs. Thus, the shape of the new options / API can be decided here. I'm certainly open to ideas.

> **Implementation examples : **

> For ObjectId :
> - I'll need to add an unknownFormats for mongodb-objectid ?
> - Then I'll need to add a new option to validateResponses as described in your solution ?
> - Do I need to configure something on request validation ?
>
> I call a request with objectId in url : /user/5fe65b0dc4a82ae0cb4d914a
> ==> With no deserialize function, I'll have a string in my route function or I'll have (and don't forget) to cast it to ObjectId in > my route function. That's what I wanted to avoid.

This is currently the case. 
The first bullet could be solved by automatically adding any new format to the `unknownFormats` list.
The second is true, given my proposed solution above.
The third, would require some registering a deserializer on the request.

The options are basically

```js
validateRequests: [{
  format: 'mongodb-objectid':
  deserializer: (s) => ObjectId(s)
}]
```

vs

```js
serdes: [{
  format: 'mongodb-objectid',
  deserializer: (s) => ObjectId(s),
}]
```

for a user with both request and response validation enabled

```js
validateRequests: [{
  format: 'mongodb-objectid',
  deserializer: (o) => o.toString()
}]

validateReponses: [{
  format: 'mongodb-objectid',
  serializer: (o) => o.toString()
}]
```

vs

```js
serdes: [{
  format: 'mongodb-objectid',
  serializer: (o) => o.toString(),
  deserializer: (s) => ObjectId(s),
}]
```

Okay, I'm with you, option 2 is better :)

cdimascio:
Another thing to consider...

for requets:

express does not deserialize requests, you get a string when a `Date` or `ObjectId` is sent as a string.

for reponses:

express does not perform any complex serialization, it calls `toJSON()/toString()`

We could start with this, that is register a custom 'object' format, then automatically add a simple deserializer that just calls `toString()`. This wouldn't require any user option and would exhibit the same behavior when validateResponses is on or off

This could be step 1. It likely solves most cases.

The we could consider improving this with customer (de)serializers

electrotype:
@pilerou I think the decision made by @cdimascio is the good one. Modifications were quite simple and they address one use case perfectly. The scope of what you suggest is larger and, probably for some, more complete.

Anyway, I'm very happy with the developments you both made, thank you again.

Since you do not want the bounty (or do you? Let me know if you changed your mind), can I ask you to pick a charity? I will send them the money and will post a proof here.

Good luck with your own use cases and with the continuation of those features!

pilerou:
Thanks for your responses. I also think you made the good solution. I work on request mecanism and other custom formats and I make a new proposition.
I don't use to contribute on github project : 
- Should I delete my fork and do a new fork from Master?
- Then should I create a new PR or continue on this one?

cdimascio:
It may be simplest to start a new fork. That said, the PR I merged in is based on your original woek, thus, you could merge master into this PR and shouldnt run into conflicts. Either way will work. Up to u

electrotype:
@pilerou Without a suggestion from you, I'm going to give the money to some charity that saves giraffes in Alaska or something.

pilerou:
Thanks @cdimascio. I do that. 
@electrotype It's very sympathetic. I really appreciate. Actually, I would feel guilty to accept money. Carmine does a wonderful job and I did a very little thing. I let you decide if you give to the last giraffes in Alaska :) or if you give the money to Carmine or if you keep the money for the next time you come in Brest (France) to drink a beer with me :) It would be a pleasure.

electrotype:
https://www.buymeacoffee.com/m97tA5c/c/631986
@cdimascio , please take care of the giraffes.
@pilerou Je te fais signe si je vais en France! Merci encore pour ton travail!

PR #501, opened 2020-12-25
add option removeAdditional to validateRequest options 

rockmagic:


cdimascio (commented):
thanks @rockmagic! this looks like a good start. 

it will be great to add some tests.

one test should set `allowUnknownQueryParameters: false` with `removeAdditional: true`. These options may need to be mutually exclusive. I'm not sure they'll work properly together. I believe this will throw as the additional properties will not have been removed by the time unknown query params are checked. this may be okay, since a user can flip `allowUnknownQueryParameters` to `true`. i would add the tests in any case

cdimascio (approved):

cdimascio:
@rockmagic apologies for the delay on this
would you mind merging latest into your branch to resolve the conflicts
i'll get this merged once its resolved

rockmagic:
@cdimascio Had to fix same `removeAdditional` type definition in `ValidateResponseOpts`, see c395537c08562ed18f2987fb294ee386018768f4 . Should I update README with details?

cdimascio (approved):

PR #504, opened 2020-12-28
fix: remove {} from OpenApiRequest.openapi type

### LINKED ISSUE #460: Type definition of OpenApiRequest has issue
In src/framework/types.ts, the definition of OpenApiRequest is rejected by typescript because of the `| {}` in its definition.  My suggestion is to remove the `| {}` altogether -- as the ? makes the openapi property optional already.  That makes it parse correctly in typescript.

Error message when referencing the .pathParams property of the OpenApiRequest variable:  ` error TS2339: Property 'pathParams' does not exist on type '{} | OpenApiRequestMetadata'`

Current code:

```
export interface OpenApiRequest extends Request {
  openapi?: OpenApiRequestMetadata | {};
}
```

Suggested change:

```
export interface OpenApiRequest extends Request {
  openapi?: OpenApiRequestMetadata;
}
```

cdimascio:
fixes #460


cdimascio:
closing in favor of #520

PR #525, opened 2021-01-27
Add validateApiDoc props in OpenApiValidatorOpts

ZhouJian26:
Hi, i added validateApiDoc props in OpenApiValidatorOpts interface. I noticed that it was missing, watching these few lines:

https://github.com/cdimascio/express-openapi-validator/blob/e57e04e446f31f4a5b11227d265a5ed010a08493/src/framework/index.ts#L34-L54

Now it is possible to disable the OpenAPI specification validation.

```ts
app.use(
  OpenApiValidator.middleware({
    apiSpec: './openapi.yaml',
    validateApiDoc: false // true by default
  }),
);
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -478,6 +478,7 @@ OpenApiValidator.middleware({
   apiSpec: './openapi.yaml',
   validateRequests: true,
   validateResponses: true,
+  validateApiSpec: true,
   validateSecurity: {
     handlers: {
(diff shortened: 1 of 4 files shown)


cdimascio (approved):
thanks @ZhouJian26! 
i modified this some to rename the new option to `validateApiSpec` for consistency with the `apiSpec` option

Issue #528, opened 2021-02-01
response 204 with no body


th3g3ntl3men:
**Describe the bug**
Having a response of 204 with no body produces an error. 204 being no content, so having a null body would be valid (I think). 
```yaml
openapi: 3.0.0
info:
  description: api
  version: '1.0.0'
paths:
  /test: 
   get: 
     ....
     responses: 
       204: 
        description: success
```

cdimascio:
@th3g3ntl3men this appears to work. the unit test passes.
you can also try it manually like so

Set up and run example 4:

```shell
cd examples/4-eov-operations-babel
npm i
npm run dev
```

curl the delete endpoint which returns `204` no content
(response validation is turned on in this example)

```shell
curl -i -X DELETE localhost:3000/v1/pets/1
HTTP/1.1 204 No Content
X-Powered-By: Express
Date: Wed, 03 Feb 2021 01:49:17 GMT
Connection: keep-alive
Keep-Alive: timeout=5

```

spec for the above operation

```
delete:
  description: deletes a single pet based on the ID supplied
  operationId: deletePet
  x-eov-operation-id: pets#delete
  x-eov-operation-handler: routes/pets
  parameters:
    - name: id
      in: path
      description: ID of pet to delete
      required: true
      schema:
        type: integer
        format: int64
  responses:
    '204':
      description: pet deleted
```

all in all, i'm not seeing the issue. are you running the latest `v4.10.9`
if you're still seeing the issue, i'm going to need some help reproducing this.

also, i cloned your repo to see where you code is executed. i don't ever see it execute in my examples. i also see the 204 no content work both when you code is present and when not.

th3g3ntl3men:
you know what, I am so sorry, I didn't  check my version. I am a little behind on versions. i updated and it appears to work. *sigh* I need to get in the habit of checking versions. sorry.

PR #529, opened 2021-02-01
if status is 204 passthrough

th3g3ntl3men:
Status Code 204 is "No Content" So this could allows for a passthrough to occur. 

If a user uses 204 and tries to return a body this is against spec, so throw an error.

[issue](https://github.com/cdimascio/express-openapi-validator/issues/528)

I am not an expert so I found the easiest path to make it work. Feel free to rip apart and help me make it better.

### DIFF
--- a/examples/4-eov-operations-babel/package-lock.json
+++ b/examples/4-eov-operations-babel/package-lock.json
@@ -5,9 +5,9 @@
   "requires": true,
   "dependencies": {
     "@apidevtools/json-schema-ref-parser": {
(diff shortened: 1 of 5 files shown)


cdimascio:
@th3g3ntl3men see my [comment](https://github.com/cdimascio/express-openapi-validator/issues/528#issuecomment-772149472) in the issue.

th3g3ntl3men:
closing this, updated to the latest and it works. Should of trusted the package :)

Issue #530, opened 2021-02-02
electron asar - failed to open spec dir


phpfs:
Hi there,

thank you very much @cdimascio for bringing this library to live!
We are currently working on a project which involves integrating an express webserver into an electron app. This means that all code and the spec file will have to be packaged using `electron-builder` before we can ship it to our clients - as a single asar archive.

**Describe the bug**
Therefore, all source files can then be read individually, but folders do not really exist anymore.
As your library attempts to change its working directory to where the spec file resides, the whole electron process crashes due to the illegal operation - i.a.w. you cannot open the spec's directory.

**To Reproduce**
Package an express server as an electron app using electron-builder and attempt to read the spec file from within the produced asar archive.

**Actual behavior** 
The electron process crashes since opening a directory in the asar file is not possible.

**Expected behavior**
The spec file is read and consumed as expected without crashing electron.

**Examples and context**
We identified the following code snippet as cause to the issue we observe. As a quick fix, simply removing line 87 resolves the issue.

`./src/framework/index.ts#L87`
```
if (fs.existsSync(absolutePath)) {
        // Get document, or throw exception on error
        try {
          **process.chdir(specDir);**
          return $refParser.mode === 'dereference'
            ? $RefParser.dereference(absolutePath)
            : $RefParser.bundle(absolutePath);
        } finally {
          process.chdir(origCwd);
        }
      } else {
        throw new Error(
          `${this.loggingPrefix}spec could not be read at ${filePath}`,
        );
      }
    }
```

Although this seems like a feasible solution to us, we do not really know the reasoning behind the code snippet and the design decisions it is based upon. How could we fix this properly? :)

Thank you very much!

cdimascio:
fixed in `v4.10.10`

PR #531, opened 2021-02-02
Fix: electron asar - failed to open spec dir

phpfs:
**This PR attempts to resolve the issue described in ticket #530 "electron asar - failed to open spec dir".**

Although the removed statements seem to have had intention, their absence does not cause any issue with the current mocha test suite. Locally, the PR runs with 337 test passed and 4 pending - exactly like beforehand. To me it makes sense that the code works without these statements since actual reading takes place using the absolute path only.

It remains to be discussed wether this is a proper fix for the issue (or just a workaround) and if this may cause issues for some users.

Thank you very much @cdimascio for your help :)

### DIFF
--- a/src/framework/index.ts
+++ b/src/framework/index.ts
@@ -87,7 +87,7 @@ export class OpenAPIFramework {
           return $refParser.mode === 'dereference'
             ? $RefParser.dereference(absolutePath)
             : $RefParser.bundle(absolutePath);
-        }
+        } finally {}
       } else {
         throw new Error(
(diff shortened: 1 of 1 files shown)


cdimascio:
@phpfs yes. it appears we no longer need to change dir. cool! thanks for this

cdimascio:
one minor nit, please remove the empty finally block. once in, i'll get this merged in and a new version rolled out asap

cdimascio (commented):
cleanup unneeded finally block and we'll get this merged in

cdimascio:
nevermind ;) i cleaned it up

Issue #541, opened 2021-02-24
validateApiSpec: false does not appear to be working


aaronluman:
running version 4.11.0 on express 4.17.1

the openapi.yaml file `openapi: 3.0.2`

setting

```
OpenApiValidator.middleware({
    apiSpec: './openapi.yaml',
    validateApiSpec: false // true by default
  })
```
does not appear to prevent validation of the spec. Running with an error in the spec (duplicate enum value) and validateApiSpec set to false triggers this error. 

```
Error: schema is invalid
```

express-openapi-validator/dist/framework/index.js:initialize is called for version 3.0.2 
the processing gets to
express-openapi-validator/node_modules/ajv/lib/ajv.js:_addSchema and willValidate = false because skipValidation is set to true
_addSchema gets called again for schema draft-04 but willValidate = true because this call does not include skipValidation = false in the method call

ajv:validateSchema then runs and fails on schema draft-04

**Expected behavior**
setting validateApiSpec should prevent ajv:validateSchema from running.

PR #543, opened 2021-02-25
propagate validateApiSpec: false to ajv

aaronluman:
resolves issue [541](https://github.com/cdimascio/express-openapi-validator/issues/541)

For context. We have a somewhat large spec file and would like to avoid the overhead of validating the spec on every request. There is also a concern that concurrent merges to our deploy could interject duplicate enum values, etc which would effectively kill the api. I created a test in our system specifically for validating the spec.

There is possibly a better way to do this and I am open to any suggestions that you might have.

For example, we could change the validateApiSpec key to an object|boolean (see [OpenApiValidatorOpts.validateResponses](https://github.com/cdimascio/express-openapi-validator/blob/master/src/framework/types.ts#L112)) so that users of this package can decide how deeply they want to disable the validation.

Also, there is probably a cleaner way to pass the validateSchema flag to ajv than to override the options on each validator.

I didn't add any new tests yet but it is not breaking any existing tests and manually verified that ajv/lib/ajv.js:validateSchema does not get called with this override

-- edit --

I updated the branch so that it allows for passing in boolean or validation suppression options and added validateSchema to the baseOptions response

### DIFF
--- a/README.md
+++ b/README.md
@@ -679,7 +679,9 @@ Determines whether the validator should validate the OpenAPI specification. Usef
 
 - `true` (**default**) - validate the OpenAPI specification.
 - `false` - do not validate the OpenAPI specification.
-
+- `{ ... }`
+  - `suppressValidation` same as boolean false
+  - `deeplySuppress` also notify dependencies to not validate
 
 ### ▪️ formats (optional)
 
--- a/src/framework/index.ts
+++ b/src/framework/index.ts
@@ -36,6 +36,7 @@ export class OpenAPIFramework {
     const validator = new OpenAPISchemaValidator({
       version: apiDoc.openapi,
       // extensions: this.apiDoc[`x-${args.name}-schema-extension`],
+      validateApiSpec,
     });
 
     if (validateApiSpec) {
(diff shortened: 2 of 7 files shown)


cdimascio:
@aaronluman thanks for the PR and tackling disabling spec validation. can you describe in detail what is being validated on every request that you do not expect? Thanks!

cdimascio:
@aaronluman please have a look at https://github.com/cdimascio/express-openapi-validator/pull/544, i borrowed a number of your ideas but removed some of the configuration.

aaronluman:
@cdimascio I think that your version handles the use case, thanks for the attention. I'll pull your branch into my project this evening after I put the kids to bed and double check but at first glance it looks like it will.

The 'per request' issue is probably not something that most users will run into. Basically, our project is too large to attach this to all of our endpoints at once, so our rollout plan is to attach it to a few endpoints at a time. While testing I found that having:

```
...
app.post('/example-one/id', postExampleOneHandler); // not ready to implement validation
...
app.get('/example-one', OpenApiValidator.middleware(config), getExampleOneHandler);
app.put('/example-one/id', OpenApiValidator.middleware(config), putExampleOneHandler);
...
```
will cause the the openapi spec file to load / parse / validate multiple times. I resolved this internally by wrapping the OpenApiValidator.middleware(config) with a singleton so that only loads once

cdimascio:
#544 incorporates key elements from this PR.
closing this out in favor of #544

PR #544, opened 2021-02-28
fix: observe validateApiSpec and avoid schema re-checks (performance)

cdimascio:
related to @aaronluman's https://github.com/cdimascio/express-openapi-validator/pull/543

fixes:
- #541
- avoid schema re-checks (perf improvement)

behavior:
- ensure `validateApiSpec` is observed. currently, setting it to `false` is ignored
- disables schema validation on subsequent validations since validation will have occurred on startup if `validateApiSpec` is `true`. 
- if  `validateApiSpec` is `false` schema validation is skipped entirely which was the initial intended behavior

### DIFF
--- a/src/framework/ajv/options.ts
+++ b/src/framework/ajv/options.ts
@@ -1,81 +1,87 @@
-import ajv = require("ajv");
-import { OpenApiValidatorOpts, Options, RequestValidatorOptions, ValidateRequestOpts, ValidateResponseOpts } from "../types";
+import ajv = require('ajv');
+import {
+  OpenApiValidatorOpts,
+  Options,
+  RequestValidatorOptions,
+  ValidateRequestOpts,
+  ValidateResponseOpts,
+} from '../types';
 
 export class AjvOptions {
-    private options: OpenApiValidatorOpts;
-    constructor(options: OpenApiValidatorOpts) {
-      this.options = options;
-    }
-    get preprocessor(): ajv.Options {
-      return this.baseOptions();
-    }
-  
-    get response(): ajv.Options {
-      const { coerceTypes, removeAdditional } = <ValidateResponseOpts>(
-        this.options.validateResponses
-      );
-      return {
-        ...this.baseOptions(),
-        useDefaults: false,
-        coerceTypes,
-        removeAdditional,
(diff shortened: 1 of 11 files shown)


src/framework/ajv/options.ts
```diff
@@ -0,0 +1,86 @@
...
+          serDesMap[serDesObject.format].serialize = serDesObject.serialize;
+        }
+        if (serDesObject.deserialize) {
+          serDesMap[serDesObject.format].deserialize = serDesObject.deserialize;
+        }
+      }
+    }
+
+    return {
+      validateSchema: false, // only needed on startup
```
aaronluman, line 69: What do you mean by "only needed on startup"? 
Setting this to always false prevents the test that i wrote in my project that I use to run a deep validation of the spec (the ajv check catches duplicate enum values, for example).

adding validateApiSpec to the const { ... } = this.options on line 48 and changing this line to `validateSchema: validateApiSpec` allows for full validation suppression when validateApiSpec is set to false but still allows for full schema validation otherwise

cdimascio: Good catch...

Validation occurs here on startup https://github.com/cdimascio/express-openapi-validator/blob/master/src/framework/openapi.schema.validator.ts#L10

that said, ajv isn’t set up the same way there as it is for request and response validation. Perhaps it should. It may make sense the startup code to create an ajv instance by using this code (but setting validateSchema true) https://github.com/cdimascio/express-openapi-validator/blob/master/src/framework/ajv/index.ts#L21

that should cause the complete validate on startup, which should make it okay to set validateSchema to false in all other cases

aaronluman: interesting. running the tests this morning everything appears to be working as expected. The startup check is catching the schema errors when I set `validateApiSpec: true`.

aaronluman (approved):

src/framework/ajv/options.ts
cdimascio: fantastic! i'll work to get this merged in. thanks for your help

cdimascio:
@all-contributors add @aaronluman for code and test

PR #549, opened 2021-03-02
Updated info about import

crudo:
There is no default import, neither named import, so this could help.


cdimascio (approved):
thanks, @crudo!

Issue #476, opened 2020-12-01
Documentation for using express-openapi-validator with NestJS


ghost:
I am new to NodeJS and TypeScript and I am writing a service with [NestJS](https://nestjs.com/).

Is there an example for how to register express-openapi-validator as a [NestJS middleware](https://docs.nestjs.com/middleware)?

bernhardvollnhofer:
I was able to register the validator with NestJS now:

```typescript
configure(consumer: MiddlewareConsumer) {
    middleware({
        apiSpec: `${__dirname}/../api-doc/bff-api.yaml`
    }).forEach(value => consumer.apply(value).forRoutes(OrganizationController))
}
```

ErolPeel:
I was able to register the validator with NestJS now:

```typescript
configure(consumer: MiddlewareConsumer) {
    middleware({
        apiSpec: `${__dirname}/../api-doc/bff-api.yaml`
    }).forEach(value => consumer.apply(value).forRoutes(OrganizationController))
}
```

I hope, this is correct.

adriengibrat:
Very interesting, could you provide a more detailed snippet or a link to an example repository.
I'd really like to have a API spec first approach for NestJS (instead of the current Swagger generation).

thanks ;)

cdimascio:
@adriengibrat, there is an [SO](https://stackoverflow.com/questions/65092351/how-to-install-express-middleware-express-openapi-validator-in-nestjs) that describes an approach similar to @ghost's.

 if it works out, please submit a PR with steps to integrate. it will be great to add them to the readme.

@ghost, it sounds like you've had some success. could you post a working example to a github repo and link it here? it will be great to have working code that folks can experiment with

ahilke:
@cdimascio I created a demo project here: https://github.com/ahilke/nestjs-with-express-openapi-validator/tree/express-openapi-validator-476

Hope it helps!

cdimascio:
Thanks @ahilke! Feel free to create a PR that's adds this code to the examples folder in this project.

ahilke:
Problem is, I haven't found a good solution for the [exception filter](https://github.com/ahilke/nestjs-with-express-openapi-validator/blob/express-openapi-validator-476/src/openapi-exception.filter.ts) yet:

```ts
import { ArgumentsHost, Catch, ExceptionFilter } from "@nestjs/common";
import { Response } from "express";

// FixMe: Use error types from index file
[...]

PR #554, opened 2021-03-05
doc: Add NestJS Example

ahilke:
Resolves https://github.com/cdimascio/express-openapi-validator/issues/476

Not sure if it should also be mentioned in the root README somewhere?


cdimascio (approved):
@ahilke thank you. this is fantastic!

PR #559, opened 2021-03-12
doc: Clean up README and Nestjs Example

ahilke:
I noticed there is a Q&A in the README that no longer seems relevant and removed it.
I also used the opportunity to fix some typos I noticed, as well as remove some vestiges of the NestJS example I forgot when trimming down the example.


cdimascio (approved):
fantastic! thank you again

PR #560, opened 2021-03-12
feat: Add Allow Header on 405

### LINKED ISSUE #467: How to add an "Allow" header to the response when a 405 error occures?
On a "method not allowed" (`405`) error, I'd like to be able to add a [Allow ](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Allow)header to the response. This header would contain the valid methods.

For example, if a path only supports `GET` requests, and  a `POST` request is received, I'd like to be able to send a `Allow: GET` header.

Do you know how to achieve that?

ahilke:
Resolves #467 

A few notes:

1. This works with the default express error handler. If you use a custom error handler, you might need to add the headers manually. 
2. The test case uses the default express error handler to showcase this. Sadly this means the test prints an error to the console:
```
express-openapi-validator
  ✓ adds allow header to 405 - Method Not Allowed
Method Not Allowed: PUT method not allowed
  at /home/ahilke/projects/express-openapi-validator/src/middlewares/openapi.metadata.ts:31:15
  at /home/ahilke/projects/express-openapi-validator/src/openapi.validator.ts:146:18
  at processTicksAndRejections (internal/process/task_queues.js:93:5)
```
3. I hope I considered all cases, as it was a tad more complex than I initially thought. Things I considered and added to the test:
    - base path (server)
    - keys under the path that are not a HTTP methods (e.g. parameters)
    - paths with parameters, so that express and open API route are different

### DIFF
--- a/examples/9-nestjs/src/filters/openapi-exception.filter.ts
+++ b/examples/9-nestjs/src/filters/openapi-exception.filter.ts
@@ -8,7 +8,7 @@ export class OpenApiExceptionFilter implements ExceptionFilter {
     const ctx = host.switchToHttp();
     const response = ctx.getResponse<Response>();
 
-    response.status(error.status).json(error);
+    response.status(error.status).header(error.headers).json(error);
   }
 }
 
@@ -22,4 +22,7 @@ interface ValidationError {
   }>;
   path?: string;
   name: string;
+  headers: {
+    [header: string]: string;
+  };
 }
--- a/src/framework/types.ts
+++ b/src/framework/types.ts
@@ -533,17 +533,23 @@ export interface ValidationErrorItem {
   error_code?: string;
 }
 
+interface ErrorHeaders {
+  Allow?: string;
+}
+
 export class HttpError extends Error implements ValidationError {
   status!: number;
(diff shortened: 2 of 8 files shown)


cdimascio (approved):
Great stuff. thanks!

cdimascio:
@ahilke I'm seeing a test failure on node 10 that is related to allow header.

Appreciate it if u can have a look. Thanks!

```
363 passing (7s)
  4 pending
  1 failing
  1) express-openapi-validator
       adds allow header to 405 - Method Not Allowed:
      AssertionError: expected { Object (x-powered-by, allow, ...) } to have property 'allow' of 'POST, GET', but got 'GET, POST'
      + expected - actual
      -GET, POST
      +POST, GET
      
      at request.put.expect.then (test/allow.header.spec.ts:26:36)
      at process._tickCallback (internal/process/next_tick.js:68:7)

```

ahilke:
@cdimascio Interesting, I was not aware that different node versions could treat key order differently. The order of the methods in the Allow header should not matter, so I adjusted the test case to ignore order in #562.

PR #564, opened 2021-03-19
feat: add req to onError handler for response validation

0xCAFEADD1C7:
Hello 👋 

First of all thanks for this project, we've been using it in my company for some time now and it's really helpful 👍 

The goal of this PR is simply to add the express request object to the `onError` handler for response validation. Our use case is that in case of error, we want to log the failing response with our logger, which is bound to the request (so that we have the request context), but I think that we can imagine a lot of useful use-cases to use the request here :)

### DIFF
--- a/README.md
+++ b/README.md
@@ -643,14 +643,15 @@ Determines whether the validator should validate responses. Also accepts respons
 
   **onError:**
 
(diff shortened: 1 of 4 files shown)


cdimascio (approved):

cdimascio:
@all-contributors add @0xCAFEADD1C7 for code and test

PR #572, opened 2021-03-30
Handle bad URI

benblack86:
I want to prevent bad data generated by a pen tester from causing a 500. For example, the following URL causes `decodeURIComponent` to throw a `URI malformed` error.

```
/accounts/%c0%ae%c0%ae%c0%af%c0%ae%c0%ae%c0%af%c0%ae%c0%ae%c0%af%c0%ae%c0%ae%c0%af%c0%ae%c0%ae%c0%af%c0%ae%c0%ae%c0%af%c0%ae%c0%ae%c0%af%c0%ae%c0%ae%c0%af%c0%ae%c0%ae%c0%af%c0%ae%c0%ae%c0%afetc%c0%afpasswd
```

Instead I want to turn it into a `BadRequest` error. Does this seem good? Happy to add a test, but not sure where to put it?

### DIFF
--- a/src/middlewares/openapi.metadata.ts
+++ b/src/middlewares/openapi.metadata.ts
@@ -3,6 +3,7 @@ import { pathToRegexp } from 'path-to-regexp';
 import { Response, NextFunction } from 'express';
 import { OpenApiContext } from '../framework/openapi.context';
 import {
+  BadRequest,
   MethodNotAllowed,
   NotFound,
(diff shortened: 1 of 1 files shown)


cdimascio (commented):
thanks. this looks good

cdimascio (approved):

cdimascio:
fixed in `v4.12.7`

PR #583, opened 2021-04-14
Name middleware

benblack86:
For tracing, anonymous functions wouldn't show up with name. I've converted these anonymous functions to named functions so it is a little more clearer what is happening. Let me know if you think a different naming schema would be better.

Screen Shot 2021-04-14 at 16 01 18

### DIFF
--- a/src/openapi.validator.ts
+++ b/src/openapi.validator.ts
@@ -119,7 +119,7 @@ export class OpenApiValidator {
     const self = this; // using named functions instead of anonymous functions to allow traces to be more useful
     let inited = false;
     // install path params
-    middlewares.push(function PathParamsMiddleware(req, res, next) {
+    middlewares.push(function pathParamsMiddleware(req, res, next) {
       return pContext
         .then(({ context, error }) => {
           // Throw if any error occurred during spec load.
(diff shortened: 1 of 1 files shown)


src/openapi.validator.ts
```diff
@@ -116,10 +116,11 @@ export class OpenApiValidator {
         };
       });
 
+    const that = this; // using named functions instead of anonymous functions to allow traces to be more useful
     let inited = false;
     // install path params
-    middlewares.push((req, res, next) =>
-      pContext
+    middlewares.push(function OpenApiValidatorPathParamsMiddleware(req, res, next) {
```
cdimascio, line 122: nit: remove the `OpenApiValidator` prefix from each named middleware function

src/openapi.validator.ts
```diff
@@ -116,10 +116,11 @@ export class OpenApiValidator {
         };
       });
 
+    const that = this; // using named functions instead of anonymous functions to allow traces to be more useful
```
cdimascio, line 119: rename `that` to `self`

cdimascio (changes requested):

cdimascio:
@benblack86 thanks!

benblack86:
Updated

src/openapi.validator.ts
```diff
@@ -116,10 +116,11 @@ export class OpenApiValidator {
         };
       });
 
+    const self = this; // using named functions instead of anonymous functions to allow traces to be more useful
     let inited = false;
     // install path params
-    middlewares.push((req, res, next) =>
-      pContext
+    middlewares.push(function PathParamsMiddleware(req, res, next) {
```
cdimascio, line 122: one more nit. use lower case `camelCase` for the function names e.g. `pathParamsMiddleware`. apologies for not being clear with my first comment

benblack86:
Updated

cdimascio (approved):
thanks!

Issue #589, opened 2021-04-26
Express res.json() returns wrong date when using open-api-validator!


babaliaris:
**Describe the bug**
When you return a JSON response that contains a date object created by an [ISO-8601](https://en.wikipedia.org/wiki/ISO_8601) string, and in the API spec (.yaml) the date field is type **Date**, you get the wrong **YYYY-MM-DD** back! For example, if the date was created using **"1996-01-03T22:00:00.000Z"** then in the response you get **1996-01-03** instead of  **1996-01-04** which is the correct **YYYY-MM-DD** representation of that UTC date. This only happens if you are using the open-api-validator, else what you get back is the actual ISO-8601 **"1996-01-03T22:00:00.000Z"**. It seems like the open-api-validator cuts the string to only keep the first 10 characters.

1) Just use the following example to test it out. You will see that the response returns **1996-01-03** as the date. 

2) Now disable the open-api-validator middleware by commenting it out. What you will get as a response will be  **"1996-01-03T22:00:00.000Z"* which of course is not the right format but it's definitely the correct DateTime!!!

**Examples and context**
```javascript
const express   = require("express");
const validator = require("express-openapi-validator");

//Create Express APP.
const app = express();

//JSON Middleware.
app.use(express.json());


//Open API Validator Middleware.
app.use(
    validator.middleware({
      apiSpec           : './api.yaml',
      validateRequests  : true,
      validateResponses : true
    })
);



//GET /user
app.get("/user", (req, res)=>
{
    const user = {
        id  : 0,
        name: "John",
        date: new Date("1996-01-03T22:00:00.000Z")
    };

    res.status(200).json(user);
});



//Open API ERROR handler.
app.use((err, req, res, next) => {
    
    res.status(err.status || 500).json({
      message: err.message,
      errors: err.errors,
    });
});


//Start Listening.
app.listen(1996, ()=>
{
    console.log("Listening!!!!");
});
```

```yaml
openapi: 3.0.0
servers:
  # Added by API Auto Mocking Plugin
  - description: SwaggerHub API Auto Mocking
    url: https://virtserver.swaggerhub.com/babaliaris/test/1.0.0
  - description: My Computer
    url: http://localhost:1996
info:
  description: This is a simple API
  version: "1.0.0"
  title: Simple Inventory API
  contact:
    email: you@your-company.com
  license:
    name: Apache 2.0
    url: 'http://www.apache.org/licenses/LICENSE-2.0.html'

paths:
[...]

PR #590, opened 2021-04-27
Created issue_589 spec file and yaml

babaliaris:
#589 
**This does not contain the fix yet, just a unit test to replicate the problem. In other words, the test should fail for now until the bug has been fixed.**

The problem should be [in this line](https://github.com/cdimascio/express-openapi-validator/blob/master/src/framework/base.serdes.ts#L16) but I wasn't able to replicate the problem in the project as I did in the distributed npm module (see #589 original post **Examples and context**).

I just created a simple test before trying to fix the bug and the result is this:
(Please note I'm in a Windows environment and I use ```npm run test:windows```)
```bash
 1) DateFormats
       should get the correct date:

      AssertionError [ERR_ASSERTION]: Expected values to be strictly equal:
+ actual - expected

+ '1996-01-03T22:00:00.000Z'
- '1996-01-04'
            ^
      + expected - actual

      -1996-01-03T22:00:00.000Z
      +1996-01-04

      at A:\NodeProjects\express-openapi-validator\test\issue_589.spec.ts:32:16
      at processTicksAndRejections (internal/process/task_queues.js:93:5)
```

The **+actual** in the assertion should be  **1996-01-03** (this is what I get while using the published npm version) and not **1996-01-03T22:00:00.000Z**
Even if I completely change the behavior of [this line](https://github.com/cdimascio/express-openapi-validator/blob/master/src/framework/base.serdes.ts#L16) the test result would still be the same which is weird.

Maybe it's not this line that causes the problem?

**Suggested Fix:**
```javascript
//d MUST be a Date Object.
const d = d.getTimezoneOffset();

//This will make sure to advance the day, month or year
//of a UTC time in such a way that the time part will be zero
//Example: "1996-01-03T22:00:00.000Z" ====> "1996-01-04T00:00:00.000Z"
d = new Date(yourDate.getTime() - (offset*60*1000));

//So now you can safely split using 'T' and keep the first part.
const correct_YYYY_MM_DD = d.toISOString().split('T')[0];
```

### DIFF
--- a/test/issue_589.spec.ts
+++ b/test/issue_589.spec.ts
@@ -1,0 +1,35 @@
+import * as path from 'path';
+import * as express from 'express';
+import * as request from 'supertest';
+import * as assert from 'assert';
+import { createApp } from './common/app';
+
+describe("DateFormats", () => {
+  let app = null;
+
+  before(async () => {
+    // Set up the express app
+    const apiSpec = path.join('test', 'resources', 'issue_589.yaml');
+    app = await createApp({ apiSpec }, 1996, (app) =>
+      app.use(
+        `${app.basePath}`,
+        express.Router().get(`/user`, (req, res) => res.json({id: 0, name: "Nick", date: new Date("1996-01-03T22:00:00.000Z")})),
+      ),
+    );
+  });
+
+  after(() => {
+    app.server.close();
+  });
+
+  it('should get the correct date', async () =>
+    request(app)
+      .get(`${app.basePath}/user`)
+      .expect(200).then((res)=>
+      {
+        assert.strictEqual(res.body.id, 0);
+        assert.strictEqual(res.body.name, "Nick"); 
+        assert.strictEqual(res.body.date, "1996-01-04"); 
+      })
+    );
+});
--- a/test/resources/issue_589.yaml
+++ b/test/resources/issue_589.yaml
@@ -1,0 +1,54 @@
+openapi: 3.0.0
+servers:
+  # Added by API Auto Mocking Plugin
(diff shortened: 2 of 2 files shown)


cdimascio:
@babaliaris this test look incorrect.
you are providing a UTC date, toISOString() is called on the data. The current behavior looks correct.

babaliaris:
> @babaliaris this test look incorrect.
> you are providing a ISO utc date, `toISOString()` is called on the date and the date portion is retained. The current behavior looks correct.

The test request [in this line](https://github.com/cdimascio/express-openapi-validator/pull/590/files#diff-f80733dd6f36eae836ba8926f270da58a87e7deecf5a7b5e404ee0979ed69a7bR16) uses a similar way as the code snippet in [issue 589](https://github.com/cdimascio/express-openapi-validator/issues/589).

Also in the YAML file [this line](https://github.com/cdimascio/express-openapi-validator/pull/590/files#diff-f7fd33b3140c1786baee0c8b1846ee9a2d1ed2bf9835414876b7d366fc1ac04aR53) I'm making sure it is a date, not DateTime. Only DateTime should return a full iso 8601 string. Currently, my test gets a full iso 8601 string while in the YAML file I described the date field as a date format so it's weird that I'm taking a full iso 8601 string in the response.

So the result should be 1996-01-03 , the same result I'm getting in [issue 589](https://github.com/cdimascio/express-openapi-validator/issues/589).

I truly can't understand why I'm getting this result. If you already know how to fix my [issue 589](https://github.com/cdimascio/express-openapi-validator/issues/589), then you are free to do it. Unfortunately, I can't figure it out. I'm really sorry.

PR #591, opened 2021-04-28
Validate no-content responses

lyndoh:
Fail validation if the api spec specifiies that a response should be empty, but the implementation has returned a body.

Includes test to cover this scenario as well as minor change to 204 'No content' using 'json()' instead of 'send()' to improve code coverage rather than simply bypassing the validation


cdimascio (approved):
@lyndoh looks good. thanks for the fix

cdimascio:
@all-contributors @lyndoh for code and test

cdimascio:
@all-contributors add @lyndoh for code and test

Issue #626, opened 2021-06-22
Operation query parameter descriptions should override path item query parameter descriptions, and not be validated in addition to them 


cd-rite:
**Describe the bug**
When an Operation parameter is defined with the same name as an applicable Path item parameter, parameters in requests are being validated  against BOTH parameter schemas.  According to the OpenAPI spec, the Operation level parameter should *override* the path item parameter, not be in addition to. 

https://swagger.io/docs/specification/describing-parameters/
https://spec.openapis.org/oas/v3.1.0#operation-object

**To Reproduce**
Create a spec with a parameter defined at the path level, and a parameter with the same name for an operation under that path. 

**Actual behavior** 
Requests are validated against BOTH definitions of the parameter, rather than just the operation one. 

**Expected behavior**
Requests will be validated against just the operation level schema, as by my reading of the openAPI Spec it should *override* the path level definition. 

**Examples and context**

This seems to be an issue with the way the applicable parameter validation schema arrays are accumulated during preprocessing.  

I think the preprocessPathLevelParameters function in the schema.preprocessor parser should only add path level parameters to the v.parameters array if there is not already an existing parameter definition by that name in the array. 

I get the behavior I expect if I change [this loop](https://github.com/cdimascio/express-openapi-validator/blob/fb5535527e2b7e9b67bf2c1e128f0196a6faf6c5/src/middlewares/parsers/schema.preprocessor.ts#L480) to this:

```
for (const param of parameters) {
    //skip adding path param if operation param already exists.
    if (!(v.parameters.some(vparam => vparam.name === param.name))){
        v.parameters.push(param);
    }
}
```

(Not sure if that's valid typescript.... I made my change in the node module)

cd-rite:
Also, just to be clear, I am talking about Query parameters specified at the path and operation levels, not parameters actually IN the path.

PR #631, opened 2021-07-07
Add Path level query param to list ONLY if there is not already an operation-level query param by the same name.

cd-rite:
Addresses my concern in issue #626 , that query params described at the Path level are not being overidden by params described at the Operation level.... Instead, parameters are being validated against BOTH param descriptions.

Apologies, but I have not been able to test this myself, I don't have any typescript experience, and keep getting compilation errors (even before my change!).


cdimascio (approved):
@cd-rite thanks for this.

cd-rite:
Awesome! Thanks for addressing the issue, and thanks for maintaining this module!

PR #632, opened 2021-07-13
Implement SerDes.jsonType option for non-object internal types.

robertjustjones:
The current SerDes assumes an object type on deserialization because the schema preprocessor hardcodes that jsonType. While arrays are objects in JS, they are distinct in JsonType so arrays will fail the object type check. 

This PR adds an option to specify an alternative like "array" jsonType and passes it through in the preprocessor code. Included are an added test and doc update.

Thanks for the work on this functionality @pilerou and @cdimascio.

### DIFF
--- a/README.md
+++ b/README.md
@@ -766,6 +766,7 @@ To create custom serializers and/or deserializers, define:
 - `format` (required) - a custom 'unknown' format that triggers the serializer and/or deserializer
 - `deserialize` (optional) - upon receiving a request, transform a string property to an object. Deserialization occurs _after_ request schema validation.
 - `serialize` (optional) - before sending a response, transform an object to string. Serialization occurs _after_ response schema validation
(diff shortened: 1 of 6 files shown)


robertjustjones:
I've since implemented a check for request data being a string before deserialization, removing skip that had been implemented if the incoming data was already of type object.  Without this check, the above jsonType="array" would allow an array to be submitted instead of enforcing the string requirement in the API Spec.  

You may want another option (strict?) for this check so I have PR'd it, but you can see the code and additional test here.

https://github.com/robertjustjones/express-openapi-validator/commit/60d34af8b90ca488ec912d609a5b6184ffdca449

cdimascio (approved):
@robertjustjones this is great. thank you

cdimascio:
@all-contributors add @robertjustjones for code and test

robertjustjones:
Thanks @cdimascio!

Issue #577, opened 2021-04-13
Exclude response validation that uses ApiExcludeEndpoint from nestjs/swagger


rienheuver:
**Related problem**
We use the response validation, but it also tries to validate responses that should not be validated. We use the `ApiExcludeEndpoint` decorator from nestjs/swagger on some endpoints to exclude it from our api-spec. However, the validator will still try to check it and then errors:
```
Cannot convert undefined or null to object
```
If I remove the decorator on the endpoint it will validate correctly again.

**Preferred solution**
Either this is a bug that should be fixed somehow, or there could be an option in response validation to exclude endpoints that have this decorator.

**Alternatives**
None yet. I can't skip the paths of these endpoints since I only want to skip certain http-methods on a certain path. Also, it would not be a good fix to manually have to add a path that is decorated with `ApiExcludeEndpoint`

cdimascio:
this error seems like a bug `Cannot convert undefined or null to object`. it should return `404`
given, can you provide an example?

i will consider `ignoreUndocumented`.

rienheuver:
Thanks for your reply!
Can't provide an example right now, our repo is too complex to share. I believe these are the steps to reproduce though:
1. generate/write api-spec for your api
2. remove a certain http-verb from an endpoint. E.g. remove POST `/users`, but keep GET `/users`. Maybe removing entire endpoint works as well, not sure
3. turn on `validateResponses`

Is that enough information?

And the `ignoreUndocumented` is quite necessary for us since we can't validate our API now because `ignorePaths` cannot distinguish between HTTP-verbs within an endpoint.

ahilke:
@rienheuver I know this issue is old, but I have a suggestion for this issue with existing code and maybe it still helps your or someone else:

If you are using a setup similar to the NestJS example, you can configure which routes the validator is applied to. [In the example, it's set to all routes](https://github.com/cdimascio/express-openapi-validator/blob/00a9c8f670186ea64657245ae2c5207c0553c778/examples/9-nestjs/src/app.module.ts#L26), but you can easily [set it to certain routes](https://docs.nestjs.com/middleware#applying-middleware) or [exclude certain routes](https://docs.nestjs.com/middleware#excluding-routes). The drawback is that you will have some amount of duplicated configuration, but at least NestJS provides a variety of ways to configure it effectively. 

It might be even worth considering extending the example to show-case this.

PR #637, opened 2021-07-26
add option to ignore undocumented #577

milo526:
Adds the `ignoreUndocumented` option as talked about in #577.

It bypasses both the 404 errors when calling routes that are not documented as well as the 405 errors when calling a route that is documented with a different method.

Feedback is very welcome 😄


cdimascio (approved):
thanks!

PR #650, opened 2021-08-27
Export types from the package root

rattrayalex:
Currently I have to do this:

```ts
import type { OpenAPIV3 } from "express-openapi-validator/dist/framework/types";
```

Looking through the `framework/types.ts` file, it seems many of the types might be useful to users of the library, though I'm not sure if that's true of all of them.

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -30,6 +30,7 @@ export const error = {
 };
 
 export * as serdes from './framework/base.serdes';
+export * from './framework/types';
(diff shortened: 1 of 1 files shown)


cdimascio:
exporting all is unlikely what we want. there is certainly value in separating out the OAS types and exporting them independently

rattrayalex:
Which types would you like to make public?

Issue #671, opened 2021-11-17
defaultResolver does not relate to default export


zzgab:
Hi! Great piece of work.

The `defaultResolver` seems to suggest that if the `operationId` symbol is not found in the exports, then it shall use `default`.

https://github.com/cdimascio/express-openapi-validator/blob/f295957b8d3356f519a1edc4fef7f0bf0002f876/src/resolvers.ts#L38

Yet, on the `return` statement, we no longer relate to a possible `default`, so the promised behaviour (to fall back to default export) does not happen.
https://github.com/cdimascio/express-openapi-validator/blob/f295957b8d3356f519a1edc4fef7f0bf0002f876/src/resolvers.ts#L47

Is there a reason for that?
Or shall I propose a PR?

Thanks

cdimascio:
@zzgab this is bug. a PR will be fantastic. thank you!

PR #675, opened 2021-11-28
#671 default export in handler

zzgab:
Hi,

issue #671 

I wasn't sure whether to add the `x-eov-operation-handler` optional property in `OperationObject` interface (since it has an important meaning in the program).

If you think it does not belong, we could instead use a `// @ts-ignore` comment in my new test before line 21.

Thanks

### DIFF
--- a/src/resolvers.ts
+++ b/src/resolvers.ts
@@ -33,18 +33,16 @@ export function defaultResolver(
     const modulePath = path.join(handlersPath, baseName);
     if (!tmpModules[modulePath]) {
       tmpModules[modulePath] = require(modulePath);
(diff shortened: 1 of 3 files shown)


cdimascio:
@zzgab thanks. This looks good. Please add the tsignore and I’ll get this merged

zzgab:
Updated.
Thanks.

cdimascio (approved):

cdimascio:
this fix is in 4.13.4. thanks again

cdimascio:
@all-contributors add @zzgab  for code and test

zzgab:
@cdimascio for all it's worth, I've fixed this JSON file here #676

PR #692, opened 2022-01-06
Run response validation middleware after operation handlers

mdwheele:
Hey @cdimascio! I'm back again 😂 

💁‍♀️  This middleware orders the operation handlers middleware _before_ the response validation middleware.

I'm sure that 90% of the time, not many users would hit this issue but here's what it can look like (based on what I'm able to pick apart).

Consider we have the following JavaScript object:

```js
const user = {
  id: 343,

  toJSON() {
    return {
      id: `${this.id}`
    }
  }
}
```

In this example, I have "native" values stored on the object and am using the `toJSON` method to present this object in accordance with my OpenAPI specification which dictates that `user.id` MUST be a `string` (not a `number`).

When I have an operation handler that looks like the following, my expectation would be that the call to `res.json` would stringify the object and _THEN_ it would be passed to eov for response validation, when configured:

```js
exports.showUser = (req, res) => {
  const user = { /* ... */ }

  res.json(user)
}
```

However this fails validation with something like `.response.id expected to be a string`. It _SHOULD_ be a string! Now, if I call `toJSON` myself before calling `res.json` everything is great.

Upon further inspection, it looks like eov validates responses before the operation handler middleware has a chance to execute. My patch simply re-orders how these are attached to the Express application. After re-ordering, I no longer have to manually call `.toJSON()` on my objects before returning a JSON response.

If helpful, I can put together a more robust example project to demonstrate reproduction. Just let me know!

### DIFF
--- a/src/openapi.validator.ts
+++ b/src/openapi.validator.ts
@@ -191,19 +191,6 @@ export class OpenApiValidator {
       });
     }
 
-    // response middleware
-    if (this.options.validateResponses) {
-      let resmw;
-      middlewares.push(function responseMiddleware(req, res, next) {
-        return pContext
-          .then(({ responseApiDoc }) => {
-            resmw = resmw || self.responseValidationMiddleware(responseApiDoc);
-            return resmw(req, res, next);
-          })
-          .catch(next);
-      })
-    }
-
     // op handler middleware
     if (this.options.operationHandlers) {
       let router: Router = null;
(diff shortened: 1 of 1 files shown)


mdwheele:
Hey @cdimascio! Just checking in to see if there's anything I can do to help out here. Would some additional tests here be useful?

mdwheele:
Hey @cdimascio! Is there anything I can do to help out here?

cdimascio:
@mdwheele I missed this. I saw you closed. I presume this is still and issue?

mdwheele:
Let me double-check that this is still an issue. I had a reproduction on one version of Express and cannot reproduce on an older version; which doesn't make much sense to me. 

If I can reproduce, I'll re-open this PR with some rest cases and we can go from there.

PR #695, opened 2022-01-14
CLS Context is lost after using multer middleware

yidongw:
related issue: https://github.com/expressjs/multer/issues/814
Used the solution described in the above link to fix the issue


sandeep3eweb:
Re-establish the CLS (Continuation Local Storage) namespace after Multer has completed the file storage process.

cdimascio (approved):

Issue #573, opened 2021-04-06
OpenAPI 3.1 support


thernstig:
**Is your feature request related to a problem? Please describe.**
OpenAPI 3.1 was released a while ago. It would be nice if it was supported.

**Describe the solution you'd like**
Full OpenAPI 3.1 support.

**Additional context**
https://openapi.tools/ lists this tool as not having OpenAPI 3.1 support.

essential-randomness:
I tried my hand at this and failed. See: https://github.com/cdimascio/express-openapi-validator/pull/653

I don't think this is possible right now, but I'm waiting for help from the AJV folks to see if they can help figure out what I'm doing wrong (or whether there is an error in any of the libraries/specs). See: https://github.com/ajv-validator/ajv/issues/1745

essential-randomness:
I'd be very happy to have my PR picked up/used as inspiration! But it seems the issues I opened in the AJV repo haven't been closed yet, so I'm unsure what changed that made 3.1 work. Happy if that's the case, and looking forward to finally being able to use this library.

seriousme:
Just a hint: [OpenApi schema validator](https://github.com/seriousme/openapi-schema-validator) works around the issue in AJV by using a modified version of the official schema in which the $dynamic refs have been replaced. (see https://github.com/seriousme/openapi-schema-validator/blob/master/test/convert-3.1.js)  This way it can still use AJV ;-)

The modified schema is available as: https://github.com/seriousme/openapi-schema-validator/blob/master/schemas/v3.1/schema.json

Hope this helps,
Hans

thernstig:
@cdimascio @JacobLey looking at #713 it says it only partly solves this, not that it solves it completely. Can you re-open?

JacobLey:
Agreed, 3.1 support should now be unblocked but is not yet implemented

Will hopefully get around to it in near future

essential-randomness:
Thank you _so_ much for your hard work! Looking forward to 3.1 support in the future!

btisdall:
Yes, many thanks Jacob.

gonenduk:
Hi,
Any updates on this?
Really appreciate your efforts here!

nexsodev:
2 years later... will nothing be done?

kennyma-eplus:
`if-then-else` schema is quite important now
hope express validator can upgrade to 3.1 soon

ahmed1hsn:
I can contribute the changes regarding this migration.

[...]

Issue #582, opened 2021-04-14
AJV version and configuration


spasquali-ww:
Love this project. Using it extensively. 

One application using it uses AJV elsewhere in the stack. For that part of the stack I'm using the latest version of AJV (8).

It seems like EOV uses AJV 6. That's fine, but wondering if a version upgrade is possible. Selfishly, I'd like to have version parity across my app.

Additionally I'd like to be able to have control over the AJV configuration used by EOV. I may have missed it but don't see where I can control the formats, keywords, and general options for AJV.

Also, would like to know/control the JSONSchema draft specification that is being supported (say, Draft 07)

Thanks!

cdimascio:
thanks @spasquali-ww, im closing this as it is effectively a duplicate of #573. moving to the latest AJV means upgrading json schema

PR #713, opened 2022-03-19
Bump AJV to v8

JacobLey:
Partially resolves https://github.com/cdimascio/express-openapi-validator/issues/573 + https://github.com/cdimascio/express-openapi-validator/issues/582

Unblocks path to 3.1 support

Appears to be fully backwards compatible (see notes about SerDes + Validation errors).

Happy to discuss changes made/justifications. I would also like to append a `prettier` commit to the end of this, but omitting for now to prevent noise. Perhaps once approved or as a follow on PR


src/framework/ajv/index.ts
```diff
@@ -1,164 +1,207 @@
...
-): Ajv.Ajv {
-  const ajv = new Ajv({
-    ...options,
-    schemaId: 'auto',
+): AjvDraft4 {
+  const { ajvFormatsMode, ...ajvOptions } = options;
+  const ajv = new AjvDraft4({
+    ...ajvOptions,
     allErrors: true,
-    meta: draftSchema,
```
JacobLey, line 30: implied by `ajv-draft-04`

src/framework/ajv/index.ts
```diff
@@ -1,164 +1,207 @@
...
     formats: { ...formats, ...options.formats },
-    unknownFormats: options.unknownFormats,
   });
+  if (ajvFormatsMode) {
+    addFormats(ajv, { mode: ajvFormatsMode });
+  }
   ajv.removeKeyword('propertyNames');
   ajv.removeKeyword('contains');
   ajv.removeKeyword('const');
+  ajv.addKeyword({
```
JacobLey, line 38: Any of these keywords without explicit validation logic will auto-resolve to true.

`path` and `components` are added because it is passed to AJV as a schema in places like this: https://github.com/cdimascio/express-openapi-validator/blob/master/src/middlewares/openapi.request.validator.ts#L274

src/framework/ajv/index.ts
```diff
@@ -1,164 +1,207 @@
...
     if (options.serDesMap) {
-      ajv.addKeyword('x-eov-serdes', {
+      ajv.addKeyword({
+        keyword: 'x-eov-serdes',
         modifying: true,
-        compile: (sch) => {
+        compile: (sch, p, it) => {
           if (sch) {
-            return function validate(data, path, obj, propName) {
+            if (sch.kind === 'res') {
```
JacobLey, line 63: See `handleSerDes`, this logic ensures that the deserializer (and equivelent serializer for res) will only run on proper schema.

Also considered splitting keyword into two forms `x-eov-req-serdes` and `x-eov-res-serdes`.

src/framework/ajv/options.ts
```diff
@@ -65,22 +60,32 @@ export class AjvOptions {
...
-      formats: this.options.formats.reduce((acc, f) => {
-        acc[f.name] = {
-          type: f.type,
-          validate: f.validate,
-        };
-        return acc;
-      }, {}),
-      serDesMap: serDesMap,
+      validateFormats: !!validateFormats,
+      formats: <Exclude<typeof formats, unknown[]>>formats,
```
JacobLey, line 73: Guaranteed to be `{}`, see `normalizeOptions`

Perhaps would benefit from a `NormalizedOptions` interface that extends `Options`?

src/framework/ajv/options.ts
```diff
@@ -65,22 +60,32 @@ export class AjvOptions {
...
-      format: validateFormats,
-      formats: this.options.formats.reduce((acc, f) => {
-        acc[f.name] = {
-          type: f.type,
-          validate: f.validate,
-        };
-        return acc;
-      }, {}),
-      serDesMap: serDesMap,
+      validateFormats: !!validateFormats,
```
JacobLey, line 72: `fast` + `full` will be truthy, then handled via `addFormats`

src/framework/openapi.schema.validator.ts
```diff
@@ -10,19 +14,26 @@ export interface OpenAPISchemaValidatorOpts {
...
     };
-
     if (!opts.validateApiSpec) {
       options.validateSchema = false;
     }
 
-    const v = new Ajv(options);
-    v.addMetaSchema(draftSchema);
+    const v = new AjvDraft4(options);
+    addFormats(v, ['email', 'regex', 'uri', 'uri-reference']);
```
JacobLey, line 36: These are the only formats referenced in spec. Can append all if desired but figured it wasn't necessary?

src/framework/types.ts
```diff
@@ -64,11 +65,17 @@ export type OperationHandlerOptions = {
   resolver: Function;
 };
 
-export type Format = {
-  name: string;
-  type?: 'number' | 'string';
-  validate: (v: any) => boolean;
-};
+export type Format =
```
JacobLey, line 68: `type` is necessary for AJV.

Functions typed with `any` are still valid (because that is how typescript works) but otherwise type must match.

Not sure if this is considered breaking... if in doubt could `default` the value to `string` and log a warning?

src/framework/types.ts
```diff
@@ -122,16 +129,20 @@ export interface OpenApiValidatorOpts {
   ignoreUndocumented?: boolean;
   securityHandlers?: SecurityHandlers;
   coerceTypes?: boolean | 'array';
+  /**
+   * @deprecated
+   * Use `formats` + `validateFormats` to ignore specified formats
+   */
   unknownFormats?: true | string[] | 'ignore';
```
JacobLey, line 136: [unknownFormats: the same can be achieved by passing true for formats that need to be ignored via `ajv.addFormat` or `formats` option.](https://ajv.js.org/v6-to-v8-migration.html)

src/framework/types.ts
```diff
@@ -122,16 +129,20 @@ export interface OpenApiValidatorOpts {
...
   serDes?: SerDes[];
-  formats?: Format[];
+  formats?: Format[] | Record<string, ajv.Format>;
```
JacobLey, line 138: This updated type better matches AJV's expected input.

I recommend deprecating the array version

src/framework/types.ts
```diff
@@ -122,16 +129,20 @@ export interface OpenApiValidatorOpts {
...
   fileUploader?: boolean | multer.Options;
   multerOpts?: multer.Options;
   $refParser?: {
     mode: 'bundle' | 'dereference';
   };
   operationHandlers?: false | string | OperationHandlerOptions;
-  validateFormats?: false | 'fast' | 'full';
+  validateFormats?: boolean | 'fast' | 'full';
```
JacobLey, line 145: `fast` and `full` should be deprecated...
perhaps a new option instead `addFormatsOptions`?

src/middlewares/openapi.request.validator.ts
```diff
@@ -191,7 +191,7 @@ export class RequestValidator {
       } else {
         throw new BadRequest({
           path: req.path,
-          message: `'${property}' should be equal to one of the allowed values: ${options
+          message: `'${property}' must be equal to one of the allowed values: ${options
```
JacobLey, line 194: I personally don't consider `should` -> `must` a breaking change.
Although multiple tests had to be updated to accommodate

The gist of the error is the same, and is "more correct" as MUST implies it is required, vs SHOULD implies recommendation (but not necessarily enforced)

src/middlewares/parsers/schema.preprocessor.ts
```diff
@@ -348,16 +352,71 @@ export class SchemaPreprocessor {
 
   private handleSerDes(
     parent: SchemaObject,
-    schema: SchemaObject,
+    schema: SchemaObject & { _serDesInternal?: boolean },
```
JacobLey, line 355: This function got a lot bigger... but it was the only way I was able to maintain backwards compatibility.

https://ajv.js.org/guide/modifying-data.html

The gist of the problem is that the `x-eov-serdes` modifies data. And AJV does not make any guarantees about the order of validations, with the exception of composite types like `allOf`

So the previous version of this function might accidentally skip validations like `format` or `pattern` because the string -> object deserialization happens first. So this new schema ensures that those validations happen, THEN deserialisation happens for requests (and vice versa for responses)

The second issue is that while the `handleSerDes` gets called for both req+res, it gets passed the same schema object. So any manipulation affects both. So `x-eov-serdes` handling was updated to always reject when `kind` does not match.

The `nullable` handling was added because AJV requires a `type` keyword alongside (which was removed from base schema). OpenAPI doesn't allow `type: 'null'` but AJV does just fine.

Biggest issue with this is that validation errors are much noisier. As there are messages for each `oneOf` failure, as well as `oneOf` itself.

test/httperror.spec.ts
```diff
@@ -7,11 +7,11 @@ describe(packageJson.name, () => {
   it('should be an instance of BadRequest', (done) => {
     console.log('Testing instaceof detection of BadRequest');
     const err = {
-      path: '.query.domain_id',
+      path: '/query.domain_id',
```
JacobLey, line 10: Similar to `must`, I don't consider the path using `/` a breaking change, despite tests needing to update to handle it

test/response.validator.spec.ts
```diff
@@ -16,7 +16,11 @@ const fakeReq: OpenApiRequest = <any>{
 };
 describe(packageJson.name, () => {
   it('should validate the using default (in this case the error object)', async () => {
-    const v = new ResponseValidator(cloneDeep(apiSpec));
+    const v = new ResponseValidator(cloneDeep(apiSpec), {
+        formats: {
+            'date-time': true,
```
JacobLey, line 21: Normally this format is added internally via `normalizeOptions`

test/serdes.spec.ts
```diff
@@ -98,7 +97,7 @@ describe('serdes', () => {
       .get(`${app.basePath}/users/1234`)
       .expect(400)
       .then((r) => {
-        expect(r.body.message).to.equal('request.params.id should match pattern "^[0-9a-fA-F]{24}$"');
+        expect(r.body.message).to.contain('request/params/id must match pattern "^[0-9a-fA-F]{24}$"');
```
JacobLey, line 100: See `handleSerDes`

Full message: `'request/params/id must match pattern "^[0-9a-fA-F]{24}$", request/params/id must pass "x-eov-serdes" keyword validation, request/params/id must be string, request/params/id must match exactly one schema in oneOf'`

JacobLey:
Re- GitGuardian

None of the referenced "secrets" look like actual secrets to me... just a `secret_identity` prop with URLs 

But @essential-randomness please let me know if those need redacting

src/framework/types.ts
JacobLey: Actually AJV allows passing options like `{ foo: x => true }` (see formats `alwaysTrue` so could use that in the case of missing `type`

essential-randomness:
> Re- GitGuardian
> 
> None of the referenced "secrets" look like actual secrets to me... just a `secret_identity` prop with URLs
> 
> But @essential-randomness please let me know if those need redacting

No actual secret there!

cdimascio:
@JacobLey thank you!
I pulled the CR and see all tests pass.
were you successfully able to bump to `v8`?

cdimascio:
@JacobLey Thank you!
ignore my previous, now deleted comment.


i pulled your PR locally and it looks solid. 
moving to the validator to ajv8 is a great step forward. i really appreciate your work.

given the size of the commit, would you be up for a chat/screenshare to walk me through the PR?

README.md
```diff
@@ -1230,6 +1230,12 @@ app.use(OpenApiValidator.middleware({
...
 
+**Q:** Can I use my own vendor extensions?
+
+**A:** Yes, any `x-*` properties are allowed (per [OpenApi Specification](https://swagger.io/docs/specification/openapi-extensions/)) and are ignored by this library.
+
+The exception is any `x-eov-*` prefixed properties, which this library reserves for its own custom vendor extensions. Some of these keywords are documented in this README, such as `x-eov-operation-id`.
```
JacobLey, line 1237: The one exception to this is `x-allow-unknown-query-parameters`. Perhaps that should be deprecated in favor of `x-eov-allow-unknown-query-parameters`?

Whoever implemented that should be ashamed 😉 
https://github.com/cdimascio/express-openapi-validator/commit/4e737b841f41d780509371afe388669d904413fb

metinogurlu:
@JacobLey some bugs fixed that many people waiting for a while in [v8.11.0](https://github.com/ajv-validator/ajv/releases/tag/v8.11.0), could you make it for this version

JacobLey:
^8.6.2 (declared version) is a caret constraint, which will support 8.11.0

npm (or whatever your package manager is) should prefer the latest version when you install it

metinogurlu:
@JacobLey ah, yes, thank you for the contribution, looking forward to the new release

src/framework/ajv/index.ts
```diff
@@ -70,56 +35,6 @@ function createAjv(
...
-  );
-  if (reservedKeywords.size > 0) {
-    throw new Error(
-      `Use of keyword(s) "${[...reservedKeywords].join(
-        '", "',
-      )}" are forbidden (reserved prefix x-eov-)`,
-    );
-  }
-  for (const customKeyword of customKeywords) {
-    ajv.addKeyword(customKeyword);
```
JacobLey: `strict: false` implies `strictSchema: false`.

Not entirely sure why I thought this was necessary... I believe there are a few tests which skip some of the options generation so perhaps I overreacted to those. But all these NOOP keywords are now ignored by default.

In the future it might be nice to allow optional `strict` options for users that want the extra enforcement, but I'll omit that for now

src/framework/types.ts
```diff
@@ -73,7 +73,6 @@ export type Format = {
 
 export type SerDes = {
   format: string;
-  jsonType?: string;
```
JacobLey, line 75: I don't think this needs any "deprecation", we just stop using it?

src/framework/ajv/index.ts
```diff
@@ -1,103 +1,124 @@
...
-    meta: draftSchema,
-    formats: { ...formats, ...options.formats },
-    unknownFormats: options.unknownFormats,
+    formats: formats,
   });
+  if (ajvFormats) {
+    addFormats(ajv, ajvFormats);
+  }
+  for (let [formatName, formatDefinition] of Object.entries(options.formats)) {
+    if (formatDefinition === '__EOV__FORMAT__ALLOW__OVERRIDE__') {
```
cdimascio, line 41: can you describe the behavior this is driving `__EOV__FORMAT__ALLOW__OVERRIDE__`

src/middlewares/util.ts
```diff
@@ -54,22 +54,40 @@ export function augmentAjvErrors(
         : e.message;
     }
   });
-  return errors;
+  const serDesPrefixes: string[] = [];
```
cdimascio, line 57: this looks to be how we clean up errors related to the injected `xEovAnyOf` used to maintain ordering for serdes. i see you use `xEovAnyOf` as an alias to anyOf. does `xEovAnyOf` have any impact on `anyOf`s that were used by clients in their actual schema?

also, what does the error look like now when `serdes` returns an error?

src/middlewares/util.ts
```diff
@@ -54,22 +54,40 @@ export function augmentAjvErrors(
...
+  errors: ErrorObject[],
 ): ValidationError {
   return {
     status,
     errors: errors.map((e) => {
       const params: any = e.params;
       const required =
-        params?.missingProperty && e.dataPath + '.' + params.missingProperty;
+        params?.missingProperty &&
+        e.instancePath + '/' + params.missingProperty;
```
cdimascio, line 85: with this change in place, do all errors consistently use `/` separators or do we still have some instances where `/` and `.` are mixed

src/framework/types.ts
cdimascio: i agreed. array version can be deprecated. can you also add this to the wiki

(note i moved the doc from README to wiki)

src/middlewares/openapi.request.validator.ts
cdimascio: agreed

src/middlewares/parsers/schema.preprocessor.ts
```diff
@@ -346,18 +350,105 @@ export class SchemaPreprocessor {
     }
   }
 
+  /**
```
cdimascio, line 353: thanks for the clear description

src/framework/ajv/options.ts
```diff
@@ -65,22 +60,21 @@ export class AjvOptions {
...
 
-    return {
-      validateSchema: false, // this is true for statup validation, thus it can be bypassed here
-      nullable: true,
+    const options: Options = {
+      strict: false,
+      strictNumbers: true,
+      strictTuples: true,
+      allowUnionTypes: false,
+      validateSchema: false, // this is true for startup validation, thus it can be bypassed here
```
cdimascio, line 68: 👍

src/middlewares/openapi.request.validator.ts
```diff
@@ -92,8 +106,8 @@ export class RequestValidator {
     });
 
     const allowUnknownQueryParameters = !!(
-      reqSchema['x-allow-unknown-query-parameters'] ??
-      this.requestOpts.allowUnknownQueryParameters
+      reqSchema['x-eov-allow-unknown-query-parameters'] ??
```
cdimascio, line 109: good call normalized on `x-eov` here and ensuring backward compat

src/middlewares/parsers/schema.preprocessor.ts
```diff
@@ -346,18 +350,105 @@ export class SchemaPreprocessor {
...
+   * The base schema is "replaced" in-place with only the `xEovAnyOf` keyword. The actual string validating
+   * schema is cloned and placed within each `anyOf`.
+   *
+   * See `augmentAjvErrors` for redaction logic.
+   *
+   * @param {object} parent - parent schema
+   * @param {object} schema - schema
+   * @param {object} state - traversal state
+   */
   private handleSerDes(
```
cdimascio, line 382: do we still add a recursive check on the response schema or were you able to remove it?

src/framework/ajv/index.ts
JacobLey: I added comment where it is being written https://github.com/cdimascio/express-openapi-validator/pull/713/commits/b8128073a3bae65d05ba5db12c24391337ea65fa#diff-45e81b10790fecfe1b0108e22869e679c99c58f5a51d8d2b35c5442f37a5cb60R402 but I can add an additional comment here (or perhaps move const to common file and comment there?)

The idea is that serDes formats need _some_ validation, whether via `true` or custom logic. In case of custom logic being provided from `ajv-formats`, this whitelisting should not turn off that logic.

We don't want custom formats to always defer to `ajv-formats` though (at least, in my opinion) so this logic has to come after `ajv-formats` and override on all other values.

src/middlewares/util.ts
JacobLey: Hmm no looks like I missed a few places. Will append another commit to handle those.

src/middlewares/util.ts
JacobLey: The error should like it used to (with exception of things like `should` -> `must`). So hides any of the gross internal logic.

See updated/reverted serdes test file https://github.com/cdimascio/express-openapi-validator/pull/713/commits/297b9919a5bc8dfd2b3c47b4040759082e499466

I think that should already be handling the various validation errors (serialization throws, pattern mismatch, format invalid...)

To my knowledge `xEovAnyOf` does not alter `anyOf` validation whatsoever, and can even exist "alongside" an `anyOf` keyword. I can add a test to prove/ensure that

cdimascio: cool, that will be nice to have such a test

src/middlewares/parsers/schema.preprocessor.ts
JacobLey: I still had it, but I just checked and it appears to be unnecessary thanks to `xEovAnyOf` (hides internal schema from traverse).

Will remove

src/framework/ajv/index.ts
JacobLey: Actually removed that constant and just applied formats in order https://github.com/cdimascio/express-openapi-validator/pull/713/commits/10e3d50354e5ad986d7a37f4de03bad9f8104993

src/middlewares/parsers/schema.preprocessor.ts
```diff
@@ -356,8 +392,20 @@ export class SchemaPreprocessor {
       !!schema.format &&
       this.serDesMap[schema.format]
     ) {
-      (<any>schema).type = [this.serDesMap[schema.format].jsonType || 'object', 'string'];
-      schema['x-eov-serdes'] = this.serDesMap[schema.format];
+      const serDes = this.serDesMap[schema.format];
```
JacobLey, line 395: Yet another serdes refactor... pretty sure that is the last time!

So the original reasoning for my complicated schema was I thought AJV did not provide any way to enforce keyword ordering (as internal keywords do not have "guaranteed" order). But custom keywords have some capability via `before` and `post`!

So the final schema looks a _lot_ more like the original, simpler version. Still some small handling had to be done for `type` + `nullable` but not nearly as gross as before.

cdimascio: This is great news, Nice find

cdimascio (commented):
this looks good. i'm considering merging shortly. is there anything else worth considering prior to doing so?

JacobLey:
I don't have anything additional to add to this PR, unless someone sees something in a PR.

`npm test` passes for me locally so I think it is functional... Happy to see it merged!

I know you mentioned _possibly_ doing a major version bump just out of caution. I'll continue to defer to you to that as it isn't _strictly_ necessary, but might be a good chance to call out "lots of internal stuff changed, use some caution". Might still be good to wait until 3.1 support to actually implement the major bump (if necessary)

cdimascio (approved):

cdimascio:
thanks @JacobLey 

the PR is now merged, you can try it out using. will make this the release version after a bit of soak time. 
```
npm install express-openapi-validator@4.14.0-beta.1
```

cdimascio:
@JacobLey i created a new branch - `v4.14.0-beta` to host this code. i'm seeing a potential performance issue. tests are running a few seconds slower.

cdimascio:
i don't see any obvious change that would cause the slowdown. clearly the internals have changed extensively with ajv8 and the external schema. that said, my expectation is that something else is at play.

the test are approximately twice as slow with these changes (`v4.14.0-beta`) to compared to 4.13.8 (on branch `mainline`) 

branches:
- `v4.14.0-beta` (new ajv8 code)
- `master` (new ajv8 code -- currently equivalent to the `v4.14.0-beta` branch)
- `mainline` (ajv6 code, 4.13.8)

cdimascio:
@JacobLey any thoughts on what might be the cause?

JacobLey:
I don't think any of the code I wrote should have significant performance implications... perhaps some of the custom keyword/serdes logic but that _ended up_ relatively unchanged

So I suspect AJV v8 is the culprit, and the internal compilation/validation logic is somehow slower.

How are you benchmarking these comparisons? Do you have a particular tool? Or just checking out the branch and seeing how long tests take to execute?

Are there particularly slow tests? Most of the example schemas are pretty simple... 2x is definitely unexpectedly poor performance from AJV. It might be good to know if the performace comes from schema "compilation" (happens once at first  request, performance only important to something like Lambda users) and "validation" (happens every request, important to everyone)

Unfortunately I can't find a good existing benchmark between AJV versions, only AJV and other validators.

cdimascio:
I haven’t dug in too deeply, thus far the it’s effectively human observation, that is I notice the mainline tests e.g. 4.13.8 complete in about 5 seconds (on my MacBook Pro) while the v4.14.0-beta.2 branch tests run in about 9 seconds.

JacobLey:
An attempt to put some hard numbers here... Adds a "listener" to test suites to track time, then reports at end (very hacky, some tests fail because they expect callbacks but those are acceptable casualities)

https://github.com/JacobLey/express-openapi-validator/tree/compile-v-validate

Hoping to:
1) Show time elapse differences
2) Highlight _compile_ vs _validate_ performance

Executed on `master` (AJV v8)
```
TOTAL SETUP TIME 106
{ compiles: 6904, validates: 2237, diff: 4667 }
```

Executed on `mainline` (AJV v6)
```
TOTAL SETUP TIME 103
{ compiles: 4378, validates: 1701, diff: 2677 }
```

Every run comes in a bit different... but those seem like good ballpark numbers. Assuming that _added_ tests between two branches is negligable... It appears compiling is ~75% slower on v8, and validation is ~20%

cdimascio:
Interesting to see that there are many more validates and diffs

cdimascio:
Some info on ajv8 perf
https://github.com/ajv-validator/ajv/issues/1386

cdimascio:
Compiles are slower in ajv8. We could offer an option to disable optimization which should gain back 30% at compile time, express openapi validator runs the compile step once when the middleware is initialized. Hence this is not an issue for long lived apis, particularly since ajv8 provides improved safety

It may also be worth investigating stand-alone mode

cdimascio:
Finally we should verify that the two validate ops are still just as performant https://github.com/cdimascio/express-openapi-validator/blob/master/src/middlewares/openapi.request.validator.ts#L174

this way we know that per req latency is just as good (or perhaps better)

JacobLey:
Standalone mode is very interesting... Would require some extra build step, but should hopefully solve most of that lambda startup time. Would probably have some sort of CLI/method that takes the OpenAPI document, and writes a file with path + method mapping to a validation function, which is then loaded by the live server.

Disabling optimization sounds like a decent compromise for lambda/cloud function users that really care about the time-to-first-validation (should be opt-in).

I tried to benchmark the per-request validation-only performance above. I was not incredibly clear, just trying to get it knocked out quickly. That was the `validates` property which did see something like a ~20% slowdown. Turning off optimization would _probably_ be worse.

I think long term we do want to adopt Ajv V8 (and future major versions). Not upgrading due to performance concerns is probably not a viable solution. Ajv still generally boasts the fastest validation times, so not any great alternatives (to my knowledge).

So this might justify putting this in a major version (already considered due to changes) bump, so we can warn that performance might take a hit, which to some _might_ be seen as a breaking change.

nd-jharn:
Any progress here? Will you be releasing this code as an update?

Issue #735, opened 2022-05-20
DoS vulnerability from dicer@0.2.5


mrded:
Hello,

Snyk is reporting a vulnerability in this repo, that is coming from the Dicer library:

```
Issues with no direct upgrade or patch:
  ✗ Denial of Service (DoS) [High Severity][https://snyk.io/vuln/SNYK-JS-DICER-2311764] in dicer@0.2.5
    introduced by express-openapi-validator@4.13.7 > multer@1.4.4 > busboy@0.2.14 > dicer@0.2.5
  No upgrade or patch available
```

Thanks

mrded:
Updating `multer > busboy@1.0.0` drops the dependency on dicer (where the vuln comes from).

ansonallard:
`multer` has an active PR for this issue: https://github.com/expressjs/multer/pull/1097

finpingvin:
> `multer` has an active PR for this issue: [expressjs/multer#1097](https://github.com/expressjs/multer/pull/1097)

They seem to have released it under 1.4.4-lts.1

cdimascio:
Fixed in 4.13.8

PR #739, opened 2022-05-30
Bump multer to version that removes dicer as sub-dependency

finpingvin:
1.4.4 version of multer has a DoS vulnerability through a sub-dependency on dicer. 1.4.4-lts.1 bumps their busboy dependency which in turn drops dicer as a dependency.

https://github.com/cdimascio/express-openapi-validator/issues/735
https://github.com/expressjs/multer/pull/1097/files#diff-7ae45ad102eab3b6d7e7896acd08c427a9b25b346470d7bc6507b6481575d519


Axenu (approved):

fgblomqvist:
Have you verified that this works as intended, per this comment: https://github.com/expressjs/multer/pull/1097#issuecomment-1140878160
?
Also, looks like you forgot to update the package-lock.json.

ghost91-:
Unfortunately, `1.4.4` is considered to be a higher semver version than `1.4.4-lts.1`. So I believe in order for this to work, the version needs to be pinned to `1.4.4-lts.1`, i.e. `1.4.4-lts.1`, instead of `^1.4.4-lts.1`. Not sure why they didn't just publish a 1.4.5, but oh well...

finpingvin:
You are right, a modified local package-lock made it so 1.4.4-lts.1 was installed. Without it I got 1.4.4.
Published with lockfile that installs 1.4.4-lts.1 but I get a bumped lockfileVersion. I don't know if that is desired. Better to pin the version?

finpingvin:
I will try to downgrade my node and see if I can get a proper lockfile without bumping lockfileVersion.

finpingvin:
With this lockfile it picks the right multer version for me

LinusU:
I've released `1.4.5-lts.1` which should be considered newer than `1.4.4`. We cannot currently release it as `1.4.5` since it contains breaking changes...

ghost91-:
> With this lockfile it picks the right multer version for me

The problem is: the lockfile is irrelevant for other packages using express-openapi-validator. The test you actually need to perform is having a package depend on your local version of express-openapi-validator, and check how the mulmtet dependency is resolved there.

That said, using the new `1.4.5-lts.1` version should hopefully do the trick, since that should be be considered newer than `1.4.4`, so when you specify it explicitly (even with `^`), npm can’t resolve that to 1.4.4 (which it does if you just specify `^1.4.4-lts.1`).

finpingvin:
> > With this lockfile it picks the right multer version for me
> 
> The problem is: the lockfile is irrelevant for other packages using express-openapi-validator. The test you actually need to perform is having a package depend on your local version of express-openapi-validator, and check how the multer dependency is resolved there.
> 
> That said, using the new `1.4.5-lts.1` version should hopefully do the trick, since that should be considered newer than `1.4.4`, so when you specify it explicitly (even with `^`), npm can’t resolve that to 1.4.4 (which it does if you just specify `^1.4.4-lts.1`).

Ah of course, stupid me! Hopefully this PR is irrelevant now anyway

finpingvin:
> > With this lockfile it picks the right multer version for me
> 
> The problem is: the lockfile is irrelevant for other packages using express-openapi-validator. The test you actually need to perform is having a package depend on your local version of express-openapi-validator, and check how the multer dependency is resolved there.
> 
> That said, using the new `1.4.5-lts.1` version should hopefully do the trick, since that should be considered newer than `1.4.4`, so when you specify it explicitly (even with `^`), npm can’t resolve that to 1.4.4 (which it does if you just specify `^1.4.4-lts.1`).

Ah of course, stupid me

cdimascio (approved):

cdimascio:
fixed in `v4.13.8`

PR #748, opened 2022-07-15
Export HTTPError

HunterKohler:
It's slightly more difficult than necessary to identity library-specific errors because the base class of all errors is not exported. Thus, `instanceof` has to be used for all error classes, or the base class recovered by reflection. In addition, the documentation should likely be changed to mention this if this proposal is suggested. I can make such modification on this branch if you wish, just reply below.

An alternative to exporting `HTTPError`, is possibly extending that from something akin to `ExpressOpenAPIValidatorError`, then exporting such a base class. That would resolve almost all naming conflicts that `HTTPError` would encounter (as it is a commonly repeated base class name in back-end node apps and libraries).

Thank all of you who have worked on the project thus-far; the library proves wildly useful for development.

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -12,6 +12,7 @@ import {
   NotFound,
   Unauthorized,
   Forbidden,
+  HTTPError,
 } from './framework/types';
 
 // export default openapiValidator;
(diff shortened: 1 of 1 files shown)


HunterKohler:
Sorry about the overly quick pull request, I just realized that the change has already been made, and I had just returned from a break thinking the import still wasn't working.

Issue #773, opened 2022-11-23
Examples with same ID causes error


HairyMike:
**Describe the bug**
Seems like examples rely on an internal 'id' field to identify them within the `express-openapi-validator` lib.  If you have multiple examples, and they share the same ID - requests to your API error out.

Example open-api doc here: https://github.com/HairyMike/bug-express-openapi-validator/blob/master/openapi.yaml

You will get a response like this:
```
Error: reference &quot;1234&quot; resolves to more than one schema
```

**To Reproduce**
```
git clone git@github.com:HairyMike/bug-express-openapi-validator.git
cd bug-express-openapi-validator.git & node index.js &
curl http://localhost:3000/v1/test
```

**Actual behavior** 
I get the following message:
```
<!DOCTYPE html>
<html lang="en">
<head>
<meta charset="utf-8">
<title>Error</title>
</head>
<body>
[...]

PR #774, opened 2022-11-23
Remove examples from apiDoc when validating requests

HairyMike:
I'd like to remove the examples from the api doc in the request validator.  I don't think they're needed while validating the request.

This could be a way to bypass this weird bug: https://github.com/cdimascio/express-openapi-validator/issues/773

### DIFF
--- a/src/middlewares/openapi.request.validator.ts
+++ b/src/middlewares/openapi.request.validator.ts
@@ -42,6 +42,8 @@ export class RequestValidator {
   ) {
     this.middlewareCache = {};
     this.apiDoc = apiDoc;
+    // Examples not needed for validation
+    delete this.apiDoc.components.examples;
     this.requestOpts.allowUnknownQueryParameters =
       options.allowUnknownQueryParameters;
     this.ajv = createRequestAjv(apiDoc, { ...options, coerceTypes: true });


josecampana:
meanwhile you merge it, I'm disabling the request validation

src/middlewares/openapi.request.validator.ts
```diff
@@ -42,6 +42,8 @@ export class RequestValidator {
   ) {
     this.middlewareCache = {};
     this.apiDoc = apiDoc;
+    // Examples not needed for validation
+    delete this.apiDoc.components.examples;
```
cdimascio, line 46: this will only remove top level examples under components. how are other handled?

dan-trewin: this also needs to be applied to `openapi.response.validator.ts`

allenrinmar:
I'm curious what the status of this PR is.  We've entered a similar issue in testing an upgrade to v5 and this looks to solve it.  Does anyone need help that I could assist with?

src/middlewares/openapi.request.validator.ts
cdimascio: Merging as is. It will be great to also add this to the response validator as suggested

cdimascio (approved):

cdimascio:
@allenrinmar will fix this shortly, have a PR here: https://github.com/cdimascio/express-openapi-validator/pull/890

PR #845, opened 2023-05-18
Remove body-parser deps in example

yutak23:
I think `body-parser` is deprecated and now we can use `express.○○○`. Therefore, I have taken the following actions in examples.

- Removed unused body-parser in package.json as well as in the sample code itself


cdimascio (approved):
Thanks!

PR #863, opened 2023-09-27
Normalize request body ContentTypes

### LINKED ISSUE #862: Request Body content-types don't consider custom parameters
**Describe the bug**
There are a few issues that come from not having the request body content types resolved by their `equivalents` instead of by sorting the params and normalizing the content-type before comparing against the available openapi content types.

**To Reproduce**/**Actual behavior**/**Examples and context**
Considering this openapi document:
```yaml
...
paths:
  /pets_content_types:
    post:
      description: Creates a new pet in the store.  Duplicates are allowed
      requestBody:
        description: Pet to add to the store
        required: true
        content:
          application/json; version=1:
            schema:
              $ref: '#/components/schemas/NewPet'
      responses:
        '200':
          description: pet response
          content:
            application/json:
              schema:
                $ref: '#/components/schemas/Pet'
        default:
          description: unexpected error
          content:
            application/json; charset=utf-8:
              schema:
                $ref: '#/components/schemas/Error'
```

If the first request has `application/json; version=1` for the request body's content type, then openapi-validator will generate a cached validator which includes the content type used.  The problem is that the cache key doesn't include the request-body params.
```
POST /pets_content_types HTTP/1.1
Content-Type: application/json; version=1
Accept: application/json

{
  "name": "curly",
  "tag": "bald"
}
```

This request would result in a validator cached as `GET-/pets_content_types-application/json`.  Subsequent requests with unsupported versions now use the same key and may return 200 instead of 415 because the requestBody content-type is in-fact not supported.

**Expected behavior**
Every request with a different requestBody content-type should be analyzed separately.  If the cache is used, then the content types must be normalized to ensure that any parameters provided in different orders create the same cache key.

All of these should also result in different cache keys so each have a chance to return a 415 status code.  Currently, they all result in the same cache key.
- application/json; version=1
- application/json; param1=1
- application/json

Likewise, both of these content-types should result in the same cache key
- application/json; version=1; charset=utf-8; param1=1
- application/json; param1=1; charset=utf-8;  version=1

rayvincent2:
Fixes #862 

First I tackled ensuring that any ContentType type could be normalized so we could perform easy comparisons between what an HTTP request may send vs what the openapi document is expecting.  By normalizing as well, I was able to easily ensure that cache keys for validators don't have any conflicts.

Tests have been added to showcase the working solution.

### DIFF
--- a/src/middlewares/util.ts
+++ b/src/middlewares/util.ts
@@ -3,23 +3,29 @@ import { Request } from 'express';
 import { ValidationError } from '../framework/types';
 
 export class ContentType {
-  public readonly contentType: string = null;
   public readonly mediaType: string = null;
-  public readonly charSet: string = null;
-  public readonly withoutBoundary: string = null;
   public readonly isWildCard: boolean;
+  public readonly parameters: { charset?: string, boundary?: string } & Record<string, string> = {};
   private constructor(contentType: string | null) {
-    this.contentType = contentType;
     if (contentType) {
-      this.withoutBoundary = contentType
-        .replace(/;\s{0,}boundary.*/, '')
-        .toLowerCase();
-      this.mediaType = this.withoutBoundary.split(';')[0].toLowerCase().trim();
-      this.charSet = this.withoutBoundary.split(';')[1]?.toLowerCase();
-      this.isWildCard = RegExp(/^[a-z]+\/\*$/).test(this.contentType);
-      if (this.charSet) {
-        this.charSet = this.charSet.toLowerCase().trim();
+      const parameterRegExp = /;\s*([^=]+)=([^;]+)/g;
+      const paramMatches = contentType.matchAll(parameterRegExp)
+      if (paramMatches) {
+        this.parameters = {};
+        for (let match of paramMatches) {
+          const key = match[1].toLowerCase();
+          let value = match[2];
+
+          if (key === 'charset') {
+            // charset parameter is case insensitive
+            // @see [rfc2046, Section 4.1.2](https://www.rfc-editor.org/rfc/rfc2046#section-4.1.2)
+            value = value.toLowerCase();
+          }
+          this.parameters[key] = value;
+        };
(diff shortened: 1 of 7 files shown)


rayvincent2:
Has anybody had a chance to review this?

src/middlewares/util.ts
```diff
@@ -30,12 +36,30 @@ export class ContentType {
...
-    if (this.charSet) {
-      return [this.mediaType, `${this.mediaType}; ${this.charSet}`];
+  public equivalents(): ContentType[] {
+    const types: ContentType[] = [];
+    if (!this.mediaType) {
+      return types;
+    }
+    types.push(new ContentType(this.mediaType));
+
+    if (!this.parameters['charset']) {
```
cdimascio, line 46: Looks like this will add utf-8 as an equivalent content type 👍

src/middlewares/util.ts
```diff
@@ -30,12 +36,30 @@ export class ContentType {
...
+      types.push(new ContentType(`${this.normalize(['charset'])}; charset=utf-8`));
     }
-    return [this.withoutBoundary, `${this.mediaType}; charset=utf-8`];
+    return types;
+  }
+
+  public normalize(excludeParams: string[] = ['boundary']) {
```
cdimascio, line 52: Nit: add the method’s return type

cdimascio:
> Has anybody had a chance to review this?

Getting to this now. Looks good. Thank you! One minor review nit

src/middlewares/util.ts
rayvincent2: Done!

rayvincent2:
> > Has anybody had a chance to review this?
> 
> Getting to this now. Looks good. Thank you! One minor review nit

Thanks for the review!  I went ahead and made the update, rebased, and squashed.  It's ready for another look.

cdimascio (approved):

PR #866, opened 2023-10-13
Reorder upload and security middlewares

### LINKED ISSUE #865: multipartMiddleware preceeds securityMiddleware so files begin uploading before security check
**Describe the bug**

When `OpenApiValidator.middleware()` is configured to apply both upload and security middleware...
```js
OpenApiValidator.middleware({
  ...
  fileUploader: {
    storage: myDiskStorageEngine,
  },
  validateSecurity: {
    handlers: {
      securityCheck: mySecurityCheck,
    },
  },
})
```
... the upload middleware (`multipartMiddleware`) runs before the security middleware (`securityMiddleware `). As a result, file uploads begin even for users that fail the security check.

The order of the middlewares is defined here:
https://github.com/cdimascio/express-openapi-validator/blob/bb8d6b80c4bc42be9213d03c46b5d3f0195d1040/src/openapi.validator.ts#L152-L179

**To Reproduce**
1. Configure `OpenApiValidator.middleware` with both a `fileUploader.storage` storage engine and a `validateSecurity` handler.
1. Attempt to upload a file using multipart/form-data with a request that does not satisfy `validateSecurity` handler.

**Actual behavior** 
File upload will be processed. For example, if the multer storage engine is a disk storage engine, the file is written to disk.

**Expected behavior**
The security check should abort the request pipeline before the file upload is processed.

**Examples and context**
I've tested reversing the order of the above mentioned middleware locally by editing `node_modules/express-openapi-validator/dist/openapi.validator.js` and the security check works as expected (file upload does not begin). I am willing to create a PR for this change, but am unsure whether there's a reason for the current order (maybe the security check for some users depends on form data?). If so, perhaps the order could be configurable?

mdmower:
- Move multipart middleware after security middleware so that security handlers can abort request pipeline before uploads are processed.

Fixes #865

### DIFF
--- a/src/openapi.validator.ts
+++ b/src/openapi.validator.ts
@@ -149,19 +149,6 @@ export class OpenApiValidator {
         .catch(next);
     });
 
-    if (this.options.fileUploader) {
-      // multipart middleware
-      let fumw;
-      middlewares.push(function multipartMiddleware(req, res, next) {
-        return pContext
-          .then(({ context: { apiDoc } }) => {
-            fumw = fumw || self.multipartMiddleware(apiDoc);
-            return fumw(req, res, next);
-          })
-          .catch(next);
-      });
-    }
-
     // security middlware
     let scmw;
     middlewares.push(function securityMiddleware(req, res, next) {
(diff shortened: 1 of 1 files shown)


mdmower:
Thanks @cdimascio, this is ready for review when you have time.

cdimascio (approved):

PR #730, opened 2022-04-05
fix: ensure JSON in form-data field is parsed IFF it is an object or array

ownagedj:
This commit fixes a problem with non primitives values in multipart/form-data field(s).

PR #878, opened 2023-11-28
Safer handling of multipart nested JSON body props

### LINKED ISSUE #875: Multer worked on 5.0.1 throws internal server error in 5.1.0
**Describe the bug**
I upgraded to v5.1.0 from v5.0.1 to get the fix where multipartMiddleware is put after securityMiddleware.
When i did this, the file is uploaded, but i get this error:

`{
    "message": "Cannot read properties of undefined (reading 'file')"
}`

And I no longer get to my requestHandler to handle req.files and furher work with the data.

**To Reproduce**
Upgrade from 5.0.1 to 5.1.0

**Actual behavior** 
File gets uploaded, but throws an internal server error that stops the request from getting to my requesthandler

**Expected behavior**
Expected the req.files to be populated with data about my uploaded files.

**Examples and context**
jest tets with supertest:
openApi def:

```
post:
  tags:
    - Files
  summary: Upload file
  description: Upload a file
  requestBody:
    required: true
    content:
      multipart/form-data:
        schema:
          oneOf:
            - title: Multiple files
              type: object
              required:
                - files
              properties:
                files:
                  type: array
                  items:
                    type: string
                    format: binary
            - title: File
              type: object
              required:
                - file
              properties:
                file:
                  type: string
                  format: binary
  responses:
    201:
      description: File successfully uploaded
      content:
        application/json:
          schema:
            type: array
            items:
              required:
                - id
              properties:
                id:
                  allOf:
                    - $ref: ../../common/schemas/id.yaml
                  description: Id of File
    default:
      $ref: ../../common/responses/defaultError.yaml
  x-eov-operation-handler: upload.controller
  x-eov-operation-id: uploadFileRequest
``` 

```typescript
        it('should upload one file and return 201 and the id of the file', async () => {
            const filePath = path.join(__dirname, '../test_file.txt');

            const response = await supertest(app)
                .post(urlUndertest)
                .set('Cookie', cookie)
                .attach('file', filePath);
            expect(response.statusCode).toBe(201);
            expect(response.body).toBeInstanceOf(Array<{ id: string }>);
        });
``` 
Request handler:
```typescript
export const uploadFileRequest = async (req: Request, res: Response, next: NextFunction) => {
    try {
        console.log(req.files);
        return res.status(201).json(await uploadFiles(req.files as Express.Multer.File[]));
    } catch (error: unknown) {
        return next(error);
    }
};
``` 

On v5.0.1

I get

```
  console.log
    [
      {
        fieldname: 'file',
        originalname: 'test_file.txt',
        encoding: '7bit',
        mimetype: 'text/plain',
        destination: '/home/thomas/tmp/uploads',
        filename: 'c5f910ddcdc142b9692668da80e94d5a',
        path: '/home/thomas/tmp/uploads/c5f910ddcdc142b9692668da80e94d5a',
        size: 36167
      }
    ]
``` 

Only changing to v5.1.0 i get

```javascript
{
    "message": "Cannot read properties of undefined (reading 'file')"
}
``` 
and of course not coming to my requestHandler, the file still uploads.

mdmower:
If a multipart request body has schema oneOf, anyOf, or allOf, then automatic parsing of JSON properties throws. An object is expected. Fix the error today and add a TODO to add support for nested JSON props in multipart requests that utilize oneOf, anyOf, or allOf.

This issue was introduced in #730.

Fixes #875

### DIFF
--- a/src/middlewares/openapi.request.validator.ts
+++ b/src/middlewares/openapi.request.validator.ts
@@ -206,7 +206,8 @@ export class RequestValidator {
   private multipartNested(req, schemaBody) {
     Object.keys(req.body).forEach((key) => {
       const value = req.body[key];
-      const type = schemaBody?.properties?.body?.properties[key]?.type;
+      // TODO: Add support for oneOf, anyOf, allOf as the body schema
+      const type = schemaBody?.properties?.body?.properties?.[key]?.type;
       if (['array', 'object'].includes(type)) {
         try {
           req.body[key] = JSON.parse(value);
@@ -214,7 +215,7 @@ export class RequestValidator {
           // NOOP
         }
       }
-    })
+    });
     return null;
   }
 


mdmower:
@cdimascio - This is a quick fix for #875. The more complete fix will be to add support for allOf, anyOf, and/or oneOf, but I didn't want to delay this bugfix while expanded schema support is worked on.

cdimascio (approved):
Thanks!

Issue #627, opened 2021-06-25
readOnly fields with data in request throw errors


Morriz:
**Describe the bug**

Request body with data for fields that are readOnly throws errors.

**To Reproduce**

Define field as readonly, and send a POST, or PUT with a payload that contains data for a readOnly field.

**Actual behavior** 

An error is thrown that the field is readOnly and should not contain data.

**Expected behavior**

The field to be discarded.

cdimascio:
thanks @Morriz, would you be interested in submitting a PR

Morriz:
Ooh, dunno. If it's within my comfort zone then yes ;) But time is not something I have lots of...

tkarls:
It would probably be a good idea to add support in the validateRequest options object to either fail the request or to remove the properties.

pilerou:
We could associate the behaviour to `removeAdditional` options ?
Then If the user post a body with data for a readonly field...
- ... and removeAdditional is set to `true`, the field is removed
- ... abd removeAddition is not set or is set to false, an error is thrown as it works today

If it's OK for you, for this behaviour, I can try to fix it

pilerou:
Other way... easyier...
Just changing ajv/index.js code only for request :
from 
```javascript
ajv.addKeyword('readOnly', {
            modifying: true,
            compile: (sch) => {
                if (sch) {
                    return function validate(data, path, obj, propName) {
                        const isValid = !(sch === true && data != null);
                        delete obj[propName];
                        validate.errors = [
                            {
                                keyword: 'readOnly',
                                schemaPath: data,
                                dataPath: path,
                                message: `is read-only`,
                                params: { readOnly: propName },
                            },
                        ];
                        return isValid;
                    };
                }
                return () => true;
            },
        });
```
to 
```javascript
ajv.addKeyword('readOnly', {
            modifying: true,
            compile: (sch) => {
                if (sch) {
                    return function validate(data, path, obj, propName) {
                        //const isValid = !(sch === true && data != null);
                        delete obj[propName];
                        return true;
                    };
                }
                return () => true;
            },
        });
```


I just tried it. It works.

sweethuman:
If this solution works will there be a PR soon?

pilerou:
Hi @sweethuman 
I just pushed in a new branch with a PR based on last revision on Master.
Unit tests are OK.

Could you review ? Or only @cdimascio can do it ?

I can make changes if needed.

sweethuman:
[...]

PR #895, opened 2024-01-30
Remove read only and write only fields

pilerou:
To meet the requirements of ticket #627, the following is proposed: 
- If ``validateRequest.removeAdditional`` is equal to ``true`` or ``'all`` or ``'failing`` then the fields received in request which are noted as ``readonly`` are removed before entering the route 
- If ``validateResponse.removeAdditional`` is equal to ``true`` or ``all`` or ``failing`` then the writeonly fields present in the response object are removed from the return flow before sending the response to the client.

ajv/index.ts had been modified to do this behaviour.
2 unit tests had been created in order to check the behaviour when ``removeAdditional : true``
Previous unit tests still works as before.

### DIFF
--- a/src/framework/ajv/index.ts
+++ b/src/framework/ajv/index.ts
@@ -117,19 +117,26 @@ function createAjv(
       compile: (sch, p, it) => {
         if (sch) {
           const validate: DataValidateFunction = (data, ctx) => {
-            const isValid = data == null;
-            if (!isValid) {
-              validate.errors = [
-                {
-                  keyword: 'readOnly',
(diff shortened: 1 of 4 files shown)


cdimascio (approved):
fantastic. thank you @pilerou !

cdimascio:
Fyi. Doc is now located here [https://cdimascio.github.io/express-openapi-validator-documentation](https://cdimascio.github.io/express-openapi-validator-documentation)

But repo here: [https://github.com/cdimascio /express-openapi-validator-documentation](https://github.com/cdimascio /express-openapi-validator-documentation)

PR #905, opened 2024-02-19
Add multipart fix when does not exist any body

hiuxmaycry:
Context in our project we have endpoints in which there is no type of body parser involved, so req.body is always undefined for those requests. This happens because the body is not consumed on the server but is simply forwarded to another API (Our API is just a proxy to another API).

Therefore, we do not validate the content of the file in any way, we only check that the endpoint exists in the SPEC.

That said, we validate that the `multipartNested` function is only applied when the body exists so as not to generate an error of `TypeError: Cannot convert undefined or null to object`.

### DIFF
--- a/src/middlewares/openapi.request.validator.ts
+++ b/src/middlewares/openapi.request.validator.ts
@@ -204,6 +204,10 @@ export class RequestValidator {
   }
 
   private multipartNested(req, schemaBody) {
+    if (!req.body) {
+      return;
+    }
+
     Object.keys(req.body).forEach((key) => {
       const value = req.body[key];
       // TODO: Add support for oneOf, anyOf, allOf as the body schema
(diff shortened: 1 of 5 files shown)


test/common/app.ts
```diff
@@ -13,19 +13,22 @@ export async function createApp(
   port = 3000,
   customRoutes = (app) => {},
   useRoutes = true,
-  apiRouter = undefined,
```
hiuxmaycry, line 16: Parameter was not used

src/middlewares/openapi.request.validator.ts
```diff
@@ -216,7 +220,6 @@ export class RequestValidator {
         }
       }
     });
-    return null;
```
hiuxmaycry, line 219: redundant return 🙂

hiuxmaycry: function return it is not used

cdimascio (approved):

hiuxmaycry:
Hi @cdimascio sorry for the inconvenience, do we have an ETA on when this fix will be released?

PR #907, opened 2024-03-05
Add `express` as peer dependency

### LINKED ISSUE #804: Running in Bazel
**Describe the bug**
Bazel complains that express is not installed. This is because express should be in "dependencies" rather than "devDependencies" in the package.json.

To fix this, "express" needs to be in "dependencies"

marvinruder:
* Fixes #804


marvinruder (commented):
If there is a minimum required version of `express` for this package to work, that version should be used instead of `*`.

PR #922, opened 2024-05-29
Support ESM import of handler module in default resolver

### LINKED ISSUE #660: Support ESM as operation handlers
**Is your feature request related to a problem? Please describe.**

It would be helpfull to support ECMAScript modules as operation handlers.

```js
export default {
  // the express handler implementation for ping
  ping: (req, res) => res.status(200).send('pong'),
};
```

**Describe the solution you'd like**

```js
operationHandlers: import.meta.url
```

**Describe alternatives you've considered**

https://github.com/cdimascio/express-openapi-validator/issues/575

### LINKED ISSUE #838: express-openapi-validator/dist/resolvers.js: require() of ES Module not supported
**Describe the bug**
I have generated a server with openapi-express-server but I need to convert it to ES6 as I will be using some libraries that do not support CommonJS anymore so I cannot use require for them.
I generated a basic server to test the refactor but I keep getting the error. I have tried with express-openapi-validator v3, v4 and v5

**To Reproduce**
Generate a server with openapi-express-server Generator and convert the files to ES6

**Actual behavior** 
require() of ES Module /controllers/InfoController.js from /node_modules/express-openapi-validator/dist/resolvers.js not supported.\nInstead change the require of InfoController.js in /node_modules/express-openapi-validator/dist/resolvers.js to a dynamic import() which is available in all CommonJS modules."

**Expected behavior**
The server should run normally

**Examples and context**
This is the Controller that is giving the error, but also all the others controllers will throw an error if used. The error appears not on the server startup but on the first API call to the REST endpoint (in this case info/version)
```javascript
// InfoController.js
import Controller from './Controller.js';
import service from '../services/InfoService.js';

const version = async (request, response) => {
  await Controller.handleRequest(request, response, service.version);
};

export default {
  version,
};
```

mdmower-csnw:
Make the default request handler resolver work in ESM projects by allowing a URL value for `options.operationHandlers`. If a URL is passed, then a dynamic import will be used to load handler modules. If a string is passed, then `require()` will be used.

Existing projects should not be adversely affected by this change. The assignment of a `URL` value to `options.operationHandlers` would be the indicator that a user wants to opt-in to this new import handling.

Fixes #660
Fixes #838

### Notice

Because this project uses `moduleResolution:node` in `tsconfig.json`, it's not possible to write `async import(...)` directly in code. It has to be obscured so that the compiler does not replace it with `require()`. Hence, there is a very obvious HACK! in this PR that obscures `import()` via `Function('x', 'return import(x)')` (it's very similar to `eval('import(...)')`).

This hack will probably sink this PR, but I figured I'd post it at least as a proof of concept. Ideally, the module resolution for tsconfig would be updated to `Node16`, but that brings its own issues since `resolveJsonModule:true` would no longer work. Fix one issue, introduce another...

### Usage example

```ts
import {fileURLToPath, pathToFileURL} from 'node:url';
import path from 'node:path';

const __dirname = path.dirname(fileURLToPath(import.meta.url));
const handlersPath = path.join(__dirname, '../path/to/my-routes');

const eov = OpenApiValidator.middleware({
  operationHandlers: pathToFileURL(handlersPath),
  // <rest of configuration>
});
app.use(eov);
```

When defining `x-eov-operation-handler` in schemas, include the file extension on the module name. For example:

```jsonc
{
  "get": {
    "summary": "Get thing",
    "x-eov-operation-id": "getThing",
    "x-eov-operation-handler": "things.js", // <-- include file extension
    "responses": {
```

### DIFF
--- a/src/framework/types.ts
+++ b/src/framework/types.ts
@@ -63,9 +63,9 @@ export type ValidateSecurityOpts = {
 };
 
 export type OperationHandlerOptions = {
-  basePath: string;
+  basePath: string | URL;
   resolver: (
-    handlersPath: string,
+    handlersPath: string | URL,
     route: RouteMetadata,
     apiDoc: OpenAPIV3.Document,
   ) => RequestHandler | Promise<RequestHandler>;
@@ -155,7 +155,7 @@ export interface OpenApiValidatorOpts {
   $refParser?: {
     mode: 'bundle' | 'dereference';
   };
-  operationHandlers?: false | string | OperationHandlerOptions;
+  operationHandlers?: false | string | URL | OperationHandlerOptions;
   validateFormats?: boolean | 'fast' | 'full';
 }
 
--- a/src/openapi.validator.ts
+++ b/src/openapi.validator.ts
@@ -56,7 +56,7 @@ export class OpenApiValidator {
     if (options.formats == null) options.formats = {};
     if (options.useRequestUrl == null) options.useRequestUrl = false;
 
-    if (typeof options.operationHandlers === 'string') {
+    if (typeof options.operationHandlers === 'string' || options.operationHandlers instanceof URL) {
       /**
        * Internally, we want to convert this to a value typed OperationHandlerOptions.
        * In this way, we can treat the value as such when we go to install (rather than
--- a/src/resolvers.ts
+++ b/src/resolvers.ts
@@ -2,40 +2,47 @@ import * as path from 'path';
 import { RequestHandler } from 'express';
 import { RouteMetadata } from './framework/openapi.spec.loader';
 import { OpenAPIV3 } from './framework/types';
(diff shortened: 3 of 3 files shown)


cdimascio (approved):
thanks

cdimascio:
#921 is now merged. please resolve conflicts and will get this in as well

mdmower-csnw:
Usage instructions could be incorporated into https://github.com/cdimascio/express-openapi-validator/tree/master/examples/5-custom-operation-resolver, but maybe the name should change to `5-operation-resolvers` with two sections: esm tips and custom resolvers? Then the link in the [documentation on operation handlers](https://cdimascio.github.io/express-openapi-validator-documentation/usage-operation-handlers/) would need to be updated as well. Let me know how I can help.

PR #926, opened 2024-06-02
Many updates - soliciting feedback

mdmower:
_This PR isn't meant to be code reviewed all at once. It's made available as a full set of commits to solicit feedback on which would be welcome. Separate PRs would be created for each approved commit._

I did a bit of repository maintenance this weekend and am interested to hear whether these commits would be welcome or if they're too heavy-handed. Notably, I thought this would be a good time to drop support for Node.js versions older than 18, allowing for tsconfig updates to use more modern module resolution strategies. For example, the hack in #922 to obscure `await import()` wouldn't be needed.

See commit descriptions for more details on each. Thanks!

P.S. The prettier formatting across the code base would cause merge conflicts with virtually every open PR. I'd recommend #882 be merged before this series, and then I'd be happy to rebase this series on the current upsteam HEAD.


mdmower:
Couple of notes:
- In commit "Update npm packages", I have a PR open at \@apidevtools/json-schema-ref-parser that would remove the need for the global type addition: https://github.com/APIDevTools/json-schema-ref-parser/pull/352
- In commit "Test improvements", I accidentally undid the tests change from https://github.com/cdimascio/express-openapi-validator/pull/854 and would restore that change.

PR #955, opened 2024-08-30
Change AJV allErrors default and support user setting

### LINKED ISSUE #954: Support configuring AJV `allErrors` to follow their recommended security practices
**Is your feature request related to a problem? Please describe.**

In [AJV's security considerations documentation](https://ajv.js.org/security.html#security-risks-of-trusted-schemas), they write "Do NOT use allErrors in production":

> Some keywords in JSON Schemas can lead to very slow validation for certain data. These keywords include (but may be not limited to):
> 
> - `pattern` and `format` for large strings - in some cases using `maxLength` can help mitigate it, but certain regular expressions can lead to exponential validation time even with relatively short strings (see [ReDoS attack](https://ajv.js.org/security.html#redos-attack)).
> - `patternProperties` for large property names - use `propertyNames` to mitigate, but some regular expressions can have exponential evaluation time as well.
> - `uniqueItems` for large non-scalar arrays - use `maxItems` to mitigate
>
> ### Do NOT use allErrors in production
> 
> The suggestions above to prevent slow validation would only work if you do NOT use `allErrors: true` in production code (using it would continue validation after validation errors).

Unfortunately, express-openapi-validator overrides whatever the user attempts to set for `allErrors`:
https://github.com/cdimascio/express-openapi-validator/blob/f20b1c9390c6d230db728f6bc528af5812f88312/src/framework/ajv/index.ts#L33-L37

Note: [`allErrors` is also set to `true` in `OpenAPISchemaValidator`](https://github.com/cdimascio/express-openapi-validator/blob/f20b1c9390c6d230db728f6bc528af5812f88312/src/framework/openapi.schema.validator.ts#L21), but that is less concerning since it is just used for OpenAPI schema validation and not end user requests. 

**Describe the solution you'd like**
It should be possible for developers to set `allErrors: false` and express-openapi-validator will respect it.

**Describe alternatives you've considered**
(none)

**Additional context**
This could help mitigate ReDOS attacks, at least to a small extent.

mdmower-csnw:
**BREAKING CHANGE**: Request and response validation stops after the first failure. Only one error will be reported even when multiple may exist. This follows best practices from AJV:
- [Security risks of trusted schemas](https://ajv.js.org/security.html#security-risks-of-trusted-schemas)
- [`allErrors` option](https://ajv.js.org/options.html#allerrors)

To report all validation errors (only recommended in development), option `allErrors` can be set in express-openapi-validator request and response validation options. For example:
```ts
app.use(
  OpenApiValidator.middleware({
    apiSpec: 'path/to/openapi.json',
    validateRequests: {
      allErrors: true,
    },
    validateResponses: {
      allErrors: true,
    },
  })
);
```

---

AJV security recommendations advise that `allErrors` should not be set to `true` in production (more details in #954). Update `createAjv()` so that it does not set `allErrors` by default and add support for letting users define `allErrors` in request and response validation options.

Added tests to ensure that the number of errors reported (when multiple are expected) is correct based on how `allErrors` is set (or unset).

The `allErrors` configuration for `OpenAPISchemaValidator` is not changed by this commit since that validation is for trusted content.

Fixes #954

### DIFF
--- a/src/framework/ajv/index.ts
+++ b/src/framework/ajv/index.ts
@@ -32,7 +32,6 @@ function createAjv(
   const { ajvFormats, ...ajvOptions } = options;
   const ajv = new AjvDraft4({
     ...ajvOptions,
-    allErrors: true,
     formats: formats,
   });
 
--- a/src/framework/ajv/options.ts
+++ b/src/framework/ajv/options.ts
@@ -45,8 +45,14 @@ export class AjvOptions {
   }
 
   private baseOptions(): Options {
-    const { coerceTypes, formats, validateFormats, serDes, allErrors, ajvFormats } =
-      this.options;
+    const {
+      coerceTypes,
+      formats,
+      validateFormats,
+      serDes,
+      allErrors,
+      ajvFormats,
+    } = this.options;
     const serDesMap = {};
     for (const serDesObject of serDes) {
       if (!serDesMap[serDesObject.format]) {
@@ -70,10 +76,10 @@ export class AjvOptions {
       coerceTypes,
       useDefaults: true,
       removeAdditional: false,
-      validateFormats: validateFormats,
+      validateFormats,
       formats,
       serDesMap,
-      allErrors: allErrors === undefined || allErrors,
+      allErrors,
       ajvFormats,
     };
 
--- a/src/framework/ajv/options.ts
+++ b/src/framework/ajv/options.ts
@@ -17,11 +17,12 @@ export class AjvOptions {
   }
 
   get response(): Options {
-    const { coerceTypes, removeAdditional } = <ValidateResponseOpts>(
+    const { allErrors, coerceTypes, removeAdditional } = <ValidateResponseOpts>(
       this.options.validateResponses
     );
     return {
       ...this.baseOptions(),
+      allErrors,
       useDefaults: false,
       coerceTypes,
       removeAdditional,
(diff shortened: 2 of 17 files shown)


mdmower-csnw:
@cdimascio - couple of questions for this PR:

1. I opted to not introduce a breaking change by defaulting `allErrors` to `true` if not defined by the end user. Do you agree with this, or would you prefer to follow AJV's recommendation and default it to `false`?
2. Should I open a 2nd PR for the [oas3.1](https://github.com/cdimascio/express-openapi-validator/tree/oas3.1) branch? Or do you plan to keep merging `master` into it over time?

cdimascio:
@mdmower-csnw thanks for highlighting this and providing a fix. the ajv default is described as:

### `allErrors`
Check all rules collecting all errors. Default is to return after the first error.

see [ref](https://ajv.js.org/options.html#verbose)

As a result, let's default to `false` following the security guidance and best practice

cdimascio:
@mdmower-csnw apologies for the multiple followups.

since this parameter `allErrors` impacts AJV behaviors, it should be available on both the request and response validator options i.e. `validateRequests` and `validateResponses`, rather than on root option object. These option objects configure each independently.

```typescript
export interface OpenApiValidatorOpts {
  apiSpec: DeepImmutable<OpenAPIV3.Document> | string;
  validateApiSpec?: boolean;
  validateResponses?: boolean | ValidateResponseOpts;  // <--- In these opts
  validateRequests?: boolean | ValidateRequestOpts;  // <--- and these opts
```

mdmower-csnw:
Thanks @cdimascio, that makes sense. I've moved the option to request and response validation options and added tests for each. A PR for related documentation is at https://github.com/cdimascio/express-openapi-validator-documentation/pull/1 .

cdimascio:
Thank you!

cdimascio (approved):

mdmower-csnw:
@cdimascio - could you check on the change history entry for this? https://github.com/cdimascio/express-openapi-validator/blob/master/CHANGE_HISTORY.md#2024-08-31 It doesn't look like the breaking change was recorded correctly.

Also, if this was released as version 5.3.4, the versions in the documentation could use an update: https://github.com/cdimascio/express-openapi-validator-documentation/pull/1#pullrequestreview-2274137954

cdimascio:
Thanks will  take a look

PR #956, opened 2024-08-30
Use lenient resolver return type

### LINKED ISSUE #952: Breaking change in signature for resolver function
**Describe the bug**
https://github.com/cdimascio/express-openapi-validator/pull/921 changed the signature of the [resolver function](https://github.com/cdimascio/express-openapi-validator/blob/master/src/framework/types.ts#L71) from

```typescript
resolver: Function;
````

to 

```typescript
resolver: (
    handlersPath: string,
    route: RouteMetadata,
    apiDoc: OpenAPIV3.Document,
  ) => RequestHandler | Promise<RequestHandler>;
```

This breaks backwards-compatibility as Express request handlers can also be arrays of request handlers.

I believe the signature should instead be

```typescript
resolver: (
    handlersPath: string,
    route: RouteMetadata,
    apiDoc: OpenAPIV3.Document,
  ) => RequestHandler | Promise<RequestHandler> | Array<RequestHandler> | Promise<Array<RequestHandler>>;
```

Unfortunately more tweaks seem to be required which I haven't found the time to investigate.

mdmower-csnw:
In #921, a stronger type was applied to `OperationHandlerOptions['resolver']` so that end users would have an idea of what the parameters are for their custom resolvers. It went too far in stipulating a return type. Set the return type to `unknown` and let users decide how much type safety they need in their resolver.

Fixes #952

### DIFF
--- a/src/framework/types.ts
+++ b/src/framework/types.ts
@@ -68,7 +68,7 @@ export type OperationHandlerOptions = {
     handlersPath: string,
     route: RouteMetadata,
     apiDoc: OpenAPIV3.Document,
-  ) => RequestHandler | Promise<RequestHandler>;
+  ) => unknown;
 };
 
 export type Format = {


mdmower-csnw:
@csuermann - could you confirm that this would fix the issue you ran into in #952? A quick way to test would be to manually edit `node_modules/express-openapi-validator/dist/framework/types.d.ts` in your project, changing
```ts
export declare type OperationHandlerOptions = {
    basePath: string;
    resolver: (handlersPath: string, route: RouteMetadata, apiDoc: OpenAPIV3.Document) => RequestHandler | Promise<RequestHandler>;
};
```
to
```ts
export declare type OperationHandlerOptions = {
    basePath: string;
    resolver: (handlersPath: string, route: RouteMetadata, apiDoc: OpenAPIV3.Document) => unknown;
};
```

When you're done testing, that file could be restored.

csuermann:
Hi @mdmower-csnw! Thank you! That seems to work!

csuermann (approved):

mdmower-csnw:
@cdimascio - this is ready for review. It fixes an issue I introduced.

cdimascio (approved):
Thanks!

Issue #803, opened 2023-01-12
Parameters declared by $ref are skipped


tom-inetum-realdolmen:
**Describe the bug**
Parameters that are declared as a $ref are skipped

**To Reproduce**
file openapi.yaml has this portion:

```yaml
openapi: 3.0.3
info:
  title: Parameter ReferenceObject
  description: test
  version: '1'
  termsOfService: https://www.google.be
  contact:
    email: noreply@test.com
security:
  - basicAuth: []
servers:
  - url: http://localhost
    description: local host
paths:
  /subject:
    get:
      summary: Test
      operationId: doTest
      parameters:
        - in: query
          name: firstParam
          schema: 
           type: string
        - $ref: "#/components/parameters/FromDateTime"
        - in: query
          name: secondParam
          schema:
            $ref: "/components/schemas/SecondParam"
      responses:
       ... (abbreviated for clarity)
components:
  schemas:
    SecondParam:
      type: string

  parameters:
    FromDateTime:
      in: query
      name: fromDateTime
      required: false
      schema: 
        type: string
```

**Actual behavior** 
parameter fromDateTime is present in `req.openapi.schema.parameters` but is not translated/parsed

**Expected behavior**
parameter fromDateTime is fully present in `req.openapi.schema.parameters`

**Examples and context**
see issue https://github.com/bug-hunters/oas3-tools/issues/59

shashanksapre:
Same. I am having to define them inline for each endpoint. This goes for both path and query parameters. :(

duncanbeevers:
This was failing for me in v5.3.1, but is working in v.5.3.3.
I believe https://github.com/cdimascio/express-openapi-validator/issues/808 also fixes this.

cdimascio:
@tom-inetum-realdolmen the provided spec appears to have errors
for instance,
`     $ref: "/components/schemas/SecondParam"` is missing the `#` 
and should be 
`     $ref: "/components/schemas/SecondParam"`

can you retry with the fix

duncanbeevers:
I'm reproducing this error locally, and have traced it to the behavior of `ajv` when validating + coercing types at https://github.com/cdimascio/express-openapi-validator/blob/master/src/middlewares/openapi.request.validator.ts#L188

I have a reproduce case, but since I think it needs fixing upstream I'm holding off on a PR.
However, I'll see if I can stick a pin in it today.

duncanbeevers:
[...]

PR #962, opened 2024-09-04
fix: Dereference path parameters

duncanbeevers:
### 🐞 The Bug
The OpenAPI spec loader has a `discoverRoutes` method which explores an OpenAPI document and gathers information about the paths and parameters used.
The list of discovered path parameters is used to install parameter-specific middleware in `src/openapi.validator.ts#installPathParams`
Path parameters declared with `$ref` were not detected in the `discoverRoutes` implementation, leading to the un-coerced values being used.

### The Fix 🔧 
By dereferencing each path parameter when building this list, we should see the same behavior for referenced path parameters and for inline path parameters.

Closes https://github.com/cdimascio/express-openapi-validator/issues/803

### DIFF
--- a/src/framework/openapi.spec.loader.ts
+++ b/src/framework/openapi.spec.loader.ts
@@ -1,3 +1,4 @@
+import { dereferenceParameter } from '../middlewares/parsers/util';
 import { OpenAPIFramework } from './index';
 import {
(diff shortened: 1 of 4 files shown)


cdimascio (approved):

cdimascio:
Thank you 👍

PR #976, opened 2024-09-12
feat(path-to-regexp): path-to-regexp 8.1.0 update

frankkeefer:
* Change path-to-regexp 6.2.2 to 8.1.0
* Update code to handled changes in path-to-regexp v6.x.x -> v8.x.x

### DIFF
--- a/src/middlewares/openapi.metadata.ts
+++ b/src/middlewares/openapi.metadata.ts
@@ -80,8 +80,6 @@ export function applyOpenApiMetadata(
       const pathKey = openApiRoute.substring((<any>methods).basePath.length);
       const schema = openApiContext.apiDoc.paths[pathKey][method.toLowerCase()];
       const _schema = responseApiDoc?.paths[pathKey][method.toLowerCase()];
-
-      let keys = [];
       const strict = !!req.app.enabled('strict routing');
       const sensitive = !!req.app.enabled('case sensitive routing');
       const pathOpts = {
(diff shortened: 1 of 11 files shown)


src/middlewares/openapi.metadata.ts
```diff
@@ -81,14 +81,16 @@ export function applyOpenApiMetadata(
...
-      const keys = [];
+      let keys = [];
       const strict = !!req.app.enabled('strict routing');
       const sensitive = !!req.app.enabled('case sensitive routing');
       const pathOpts = {
         sensitive,
         strict,
       };
-      const regexp = pathToRegexp(expressRoute, keys, pathOpts);
+      const regexpObj = pathToRegexp(expressRoute, pathOpts);
```
cvchauhan, line 91: we can remove let keys = [];
and directly use 
 const { regexpObj, keys } = pathToRegexp(expressRoute, pathOpts);

cvchauhan (changes requested):
we can remove let keys = []; and keys = regexpObj.keys;
and directly use
const { regexpObj, keys } = pathToRegexp(expressRoute, pathOpts);

src/middlewares/openapi.metadata.ts
frankkeefer: Added in requested changes. 
* removed let keys =[] and related uses
* removed const regexp = regepObj.regexp; and related uses
All current unit tests still pass locally

cvchauhan (approved):

cdimascio (approved):
thank you. this is great

dogmatic69:
Please create a release 🙏🏻

PR #977, opened 2024-09-12
fix: bump patch version to ensure latest dependencies loading

alexpit:
Would you be so kind to accept a patch update to allow the newly updated dependencies (`path-to-regexp v6.3.0`) to be loaded by the consumers of the `express-openapi-validator` module. Thanks


cvchauhan:
I think already published this version today

alexpit:
> I think already published this version today

Great. Thanks!

PR #1031, opened 2024-12-27
re-enable keywords now properly supported

uwinkelvos:
With https://github.com/cdimascio/express-openapi-validator/pull/882 merged, some keywords that where previously not relevant are should now be checked by ajv.

### DIFF
--- a/src/framework/ajv/index.ts
+++ b/src/framework/ajv/index.ts
@@ -61,9 +61,6 @@ function createAjv(
   for (let [formatName, formatDefinition] of Object.entries(options.formats)) {
     ajv.addFormat(formatName, formatDefinition);
   }
-  ajv.removeKeyword('propertyNames');
-  ajv.removeKeyword('contains');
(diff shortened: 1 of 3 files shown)


src/framework/ajv/index.ts
```diff
@@ -61,9 +61,6 @@ function createAjv(
   for (let [formatName, formatDefinition] of Object.entries(options.formats)) {
     ajv.addFormat(formatName, formatDefinition);
   }
-  ajv.removeKeyword('propertyNames');
```
cdimascio, line 64: won't these still be necessary for clients using openapi 3.0.x?

uwinkelvos: When I try to load an openapi 3.0.x spec, that uses e.g. `const`, I get a validation at middleware construction time. And even if this was bypassed somehow, Ajv would understand the keywords. So imho there is no harm at allowing them at this level (Ajv).

cdimascio (approved):

Issue #969, opened 2024-09-11
Express 5 req.query immutable


MarcelHoogesteger:
The request parameter mutator is broken when express 5 is used

When you have a query parameter definition for array values, the req.query[key] could not be updated because it is immutable in express 5.

The req.query key values are not parsed by its query parameter definitions in the OpenApi 3 specs

The req.query parameters should be parsed like defined in the OpenApi 3 schema for the query parameter

Example sort query parameter definition:
name: sort
in: query
description: Sort parameter
required: false
style: pipeDelimited
explode: false
schema:
  type: array
  items:
    type: string

req.query = { sort: "name|date"} should be having an array value after parsing: req.query = { sort: ["name", "date"]}, but it remains the same value req.query = { sort: "name|date"}

jacobshirley:
+1.

I've implemented an ugly workaround for it by making query mutable:

```typescript
app.use((req, res, next) => {
        if (req.query)
            Object.defineProperty(req, 'query', {
                writable: true,
                value: { ...req.query },
            })

        next()
 })
```

SophieDel:
+1

I am facing the same problem. I have remarked that req.query does not inherit from Object.prototype anymore ([Object: null prototype]), I dunno whether or not it can be useful to our problem.

Is it planned to fix this bug...?

Thank you !

cdimascio:
thanks all. i've started a new branch [express-5-support](https://github.com/cdimascio/express-openapi-validator/tree/express-5-support) to tackle support for express 5. Without any changes 60+ tests fails. with @jacobshirley workaround in place, 29 tests now fail. 

seeking the communities help to dig into the remaining issues

SF97:
Hello

With https://github.com/cdimascio/express-openapi-validator/pull/1036 the number goes down to 15, due to a fix I've made in some tests.

With the routing syntax changes, I don't think there's an easy way to have tests that guarantee both express 4 and express 5 compatibility, but we can think of that after all the tests route have its matching syntax fixed. Right now, I still haven't found a way to have the same wildcard tests for express 5

matas-ramanauskas-revel:
+1.
Any updates on this?

PR #1036, opened 2024-12-30
test(express-5): change routes in tests to new path route syntax

SF97:
As mentioned in https://github.com/cdimascio/express-openapi-validator/issues/969, there are several failing tests when testing this lib against express 5. One of the reasons some tests are failing is due to the fact that express 5 changed its routing syntax, which made some tests fail in the "beforeAll" hook. This MR fixes some of them

There are still some tests that use wildcard syntax that break, but I haven't found a way to get around them just yet. I'll continue to investigate it, but it can be merged in another PR


cdimascio:
@SF97 thank you, express-support-5 has only 19 errors remaining to resolve

PR #1040, opened 2025-01-13
fix: Remove all OpenAPI examples from spec before validating

### LINKED ISSUE #931: Exclude OpenAPI Examples from Response Validation
**Describe the bug**

When response validation is enabled, and my spec contains several examples with a field called `id` containing the same string value, the `express-openapi-validator` in version 5.2.0 produces an unexpected error. This is not the case for request validation.

OpenAPI examples are removed from the `apiDoc` before request validation, but not before response validation. Some weird behavior in `ajv` seems to trigger the error when compiling the schema.

It probably makes sense to remove examples before response validation as well.

**To Reproduce**

Use the following OpenAPI spec with **version 5.2.0** of the`express-openapi-validator` (this isn't a problem in 4.8.x):

```
openapi: '3.0.0'

info:
  version: 1.0.0
  title: Validation Error
  description: A test spec that triggers an validation error on identical id fields in examples.

servers:
  - url: http://{base_url}/{env_id}
    variables:
      base_url:
        default: my.app.com
        description: server
      env_id:
        default: v1
        description: path selector
paths:
  /ping:
    post:
      description: |
        ping then pong!
      operationId: ping
      responses:
        '200':
          description: OK
          content:
            application/json:
              schema:
                $ref: "#/components/schemas/Response"
              examples:
                response1:
                  summary: "Response 1"
                  value:
                    id: 'Some_ID'
                    message: 'Message 1'
                response2:
                  summary: "Response 2"
                  value:
                    id: 'Some_ID'
                    message: 'Message 2'

components:
  schemas:
    Response:
      required:
        - id
        - message
      properties:
        id:
          type: string
        message:
          type: string
```

Invoke response validation as follows:

```
const apiSpec = path.join(__dirname, 'openapi.yml');

app.use(
  OpenApiValidator.middleware({
    apiSpec,
    validateResponses: true
  }),
);
```

Setting `validateResponse` to `false` and `validateRequest` to `true` will not produce the error.

**Actual behavior** 
the `express-openapi-validator` produces the following error:

**Error Message:** `"reference \"Some_ID\" resolves to more than one schema"` (for full stack trace, see below)

**Expected behavior**
Validation of the response, without producing the described error.

**Examples and context**

Version of `express-openapi-validator`: 5.2.0

It seems like [`ajv` currently suffers from a compatibility issue](https://github.com/stoplightio/spectral/issues/2081) regarding properties named `id`, where the `id` property is used to differentiate JSON schemas generated by `avj`. If a property `id` only defines a new sub-schema, this is no problem. However, when populated with a concrete string value, two schemas having the same value will collide during compilation, resulting in the following error when validating responses:

```
"error": {
    "message": "reference \"<ID_STRING>\" resolves to more than one schema",
    "stack": [
      "Error: reference \"<ID_STRING>\" resolves to more than one schema",
      "    at ambiguos (/Users/usr/project/node_modules/ajv/lib/compile/resolve.ts:147:12)",
      "    at Ajv.addRef (/Users/usr/project/node_modules/ajv/lib/compile/resolve.ts:115:38)",
      "    at /Users/usr/project/node_modules/ajv/lib/compile/resolve.ts:106:59",
      "    at _traverse (/Users/usr/project/node_modules/json-schema-traverse/index.js:69:5)",
      "    at _traverse (/Users/usr/project/node_modules/json-schema-traverse/index.js:83:9)",
      "    at _traverse (/Users/usr/project/node_modules/json-schema-traverse/index.js:83:9)",
      "    at _traverse (/Users/usr/project/node_modules/json-schema-traverse/index.js:83:9)",
      "    at _traverse (/Users/usr/project/node_modules/json-schema-traverse/index.js:83:9)",
      "    at _traverse (/Users/usr/project/node_modules/json-schema-traverse/index.js:83:9)",
      "    at module.exports (/Users/usr/project/node_modules/json-schema-traverse/index.js:14:3)",
      "    at Ajv.getSchemaRefs (/Users/usr/project/node_modules/ajv/lib/compile/resolve.ts:102:3)",
      "    at Ajv._addSchema (/Users/usr/project/node_modules/ajv/lib/core.ts:713:37)",
      "    at Ajv.compile (/Users/usr/project/node_modules/ajv/lib/core.ts:384:22)",
      "    at ResponseValidator.buildValidators (/Users/usr/project/node_modules/@xxx/api-validation/node_modules/express-openapi-validator/src/middlewares/openapi.response.validator.ts:302:8)",
      "    at ResponseValidator._getOrBuildValidator (/Users/usr/project/node_modules/@xxx/api-validation/node_modules/express-openapi-validator/src/middlewares/openapi.response.validator.ts:113:12)",
      "    at /Users/usr/project/node_modules/@xxx/api-validation/node_modules/express-openapi-validator/src/middlewares/openapi.response.validator.ts:64:26",
    ]
  }
```

Luckily, it is rather rare that schema properties are defined with a fixed string. However, this is _not_ true for schema **examples**. In our case, we had several examples using the same value for a property called `id`. This resulted in the above error when performing **response validation**, but not request validation. Digging in the code, it turns out examples are actually removed from the generated schema in the `RequestValidator`, but _not_ for the `ResponseValidator` (see [this line](https://github.com/cdimascio/express-openapi-validator/blob/093bd3cc6cfa5716e640172b25bd0b100d3dc8ab/src/middlewares/openapi.request.validator.ts#L46) in the `RequestValidator` vs [this function](https://github.com/cdimascio/express-openapi-validator/blob/093bd3cc6cfa5716e640172b25bd0b100d3dc8ab/src/middlewares/openapi.response.validator.ts#L37) in the `ResponseValidator`).

As examples should never matter in the validation of responses, we suggest to remove examples in this case as well. In fact, probably removing the examples earlier, e.g., in [this function](https://github.com/cdimascio/express-openapi-validator/blob/093bd3cc6cfa5716e640172b25bd0b100d3dc8ab/src/framework/openapi.spec.loader.ts#L51), would be a good idea, if there is no use case I am missing.

If desired, I can create a pull request for this, but I am not sure if I will be able to take all edge cases into account.

max-at-silverflow:
This PR was made to resolve issue #931. It introduces the following changes:

Commit  189d6249b43b2de73ddb026a8619cf6d9d827629:
- Add a `stripExamples` function that takes an `OpenAPIV3.DocumentV3` or `OpenAPIV3.DocumentV3_1`, removing all `example` and `examples` properties. (**Note:** This only affects examples provided as part of the spec, not user-defined properties called `example` or `examples`).
- Call `stripExamples` in the spec loader to ensure examples are always discarded and do not affect validation in any way.
- Add minimal test setup that verifies this works. (**Note:** If you remove the `stripExamples` call, this test fails).

These changes are necessary to avoid the validator from erroring out when providing specs that include multiple examples with the same value for a property `id`. This is explained in more detail in issue #931, though the proposed solution does not cover all cases. This PR expands on the suggestions and strips all examples, instead of just the component examples.

**Note**:
Originally, this PR also ran `prettier -c -w ./` to improve the code style of the repository (49ffda64ca8952daa7511ce1f3b05932d6771170). But I realized that were way too many changes, so I reverted them in 8dfdb94a3807c93df71a84bb3adcbf14aeb42138. Sorry for the back-and-forth.

Closes #931


max-at-silverflow:
If preferred, I can split the `pretter` fixes in a separate PR as well! I just thought this way it might be more efficient.

cdimascio:
thanks for the PR @max-at-silverflow. currently, schema.preporcessor traverses the schema. as part of this process it attempts to remove examples (the logic is in schemaVisitor which calls removeExamples). given this PR, i suspect that is not actually remove all examples. can this PR be updated to leverage/fix the existing process. Note that the existing process attempts to clean up examples through a single schema traversal. this logic appears to add a second traversal. i'd like to ensure the single pass is maintained for performance.

Pointers regarding the current remove examples logic:

currently, the schema.preprocessor.ts exposes an entry point method `preProcess`. `preProcess` ultimately calls `traverseSchemas`, which provides a visitor (schemaVisitor):

```typescript
this.traverseSchemas(schemaNodes, (parent, schema, opts) =>
        this.schemaVisitor(parent, schema, opts),
      );
```

schema visitor is called during the schema walk. This visitor includes a call to `removeExamples`. ideally, this call path removes all `example` and `examples`. 

the `schemaVisitor` calls `removeExamples`:

e.g.
```typescript
private schemaVisitor(
    parent: SchemaObjectNode,
    node: SchemaObjectNode,
    opts: TraversalStates,
  ) {
    const pschemas = [parent?.schema];
    const nschemas = [node.schema];
   // ...
     // visit the node in both the request and response schema
    for (let i = 0; i < nschemas.length; i++) {
      // ...
      this.removeExamples(pschema, nschema, options)
      //..
```

ideally, this PR leverages this call path

max-at-silverflow:
Hey @cdimascio! Thanks for the response, that makes a lot of sense! I will see that I can implement it in a compatible way!

cdimascio:
@max-at-silverflow fyi, i pushed the following change which should help with some example(s) removal
see https://github.com/cdimascio/express-openapi-validator/commit/bd9426c

max-at-silverflow:
Thanks a lot for the changes! This is a great start!

Unfortunately, it does not resolve all my issue yet. In particular, duplicated ids in the `components` examples still seem to cause issues (as simulated by the `ignore.examples.spec.ts` test file).

I am currently trying to understand the pre-processor, hoping to improve the example removal even further. I will see if there is a reasonable way to achieve this.

But thanks for your efforts!

cdimascio:
No problem. Thank for continuing to investigate.

max-at-silverflow:
Alright, I made quite some progress on a more complete fix, I will update
this MR in the next few days!

cdimascio:
closing this out given the variety of related fixes delivered.
there is https://github.com/cdimascio/express-openapi-validator/pull/1051 which contains a fix, but requires more vetting given its size

PR #1048, opened 2025-02-28
Remove more examples from schemas

jstasiak:
[1] did some work to resolve a false positive schema validation issue[2]. It's about examples with the "id" property.

As reported in [3] there may be examples with such property for schemas where type is not object.

This patch hopefully fixes all remaining instances of the problem.

[1] 0ad49ec78e0a ("remove examples from schema (#890)")
[2] https://github.com/cdimascio/express-openapi-validator/issues/773
[3] https://github.com/cdimascio/express-openapi-validator/pull/890#issuecomment-1914299643


jstasiak:
Actually I just saw that there already is a PR for this, #1040, so I'll close this one.

PR #1051, opened 2025-03-05
feat: Re-design of the Preprocessor to Reach and Strip all Examples in the Schema

### LINKED ISSUE #931: Exclude OpenAPI Examples from Response Validation

max-at-silverflow:
This MR is a spin-off of #1040, and aims to fix #931. It introduces the following changes:

- Add a test to verify #931 has been fixed.
- Expose an additional type from the OpenAPIV3 namespace (and fixes `prettier` in the types file).
- Redesign the preprocessor to scan the entire API schema for examples, using the same pattern as the original implementation.

The changes are quite extensive. The new preprocessor serves as a drop-in replacement, but changes how the schema is traversed. Most importantly, it traverses much more of the schema than the old design, which is required to reach all examples. In turn, it only requires a single pass to perform pre-processing and stripping all examples, in contrast to the simpler solution provided in #1040.

If you have any questions about it, let me know!

Closes #931.
Replaces #1040.

### DIFF
--- a/src/framework/types.ts
+++ b/src/framework/types.ts
@@ -7,7 +7,7 @@ import AjvDraft4 from 'ajv-draft-04';
 import Ajv2020 from 'ajv/dist/2020';
 export { OpenAPIFrameworkArgs };
 
-export type AjvInstance = AjvDraft4 | Ajv2020 
+export type AjvInstance = AjvDraft4 | Ajv2020;
 
 export type BodySchema =
   | OpenAPIV3.ReferenceObject
@@ -48,7 +48,7 @@ export interface Options extends ajv.Options {
   ajvFormats?: FormatsPluginOptions;
 }
 
-export interface RequestValidatorOptions extends Options, ValidateRequestOpts { }
+export interface RequestValidatorOptions extends Options, ValidateRequestOpts {}
 
 export type ValidateRequestOpts = {
   /**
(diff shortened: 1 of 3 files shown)


max-at-silverflow:
I also just tested this with a larger project of my own, and with the relatively complex schema there I did not find any issues. Doesn't mean that all edge cases have been covered, but should be a good indication.

cdimascio:
@max-at-silverflow this is a large change. What is your confidence, particularly regarding the series work?

max-at-silverflow:
Yeah, it is quite a chunk... So, from my perspective, I can say the following:

- It does work with all the current tests. Depending on how confident you are into those, that should be a good indication.
- I have tested it with one of my projects, which is a proper real-world application with schemas that are way more complicated than anything in the tests.

Neither of this is conclusive, obviously. And considering the user base of this project, I would rather not promise that it is perfect and bug-free. I have spent significant effort to make it as stable as possible, but I have done this change on my own, so it was only a single pair of eyes.

If I had to make a decision here, I would probably opt for a beta release, if you have that option. That would give the opportunity to gather some more feedback in the wild.

And if anyone else wants to give their feedback, that is obviously also more than welcome!

Let me know what you think!

PR #1062, opened 2025-05-08
Optimise Ajv cache usage

0bex0:
Introducing a key which can be used to for caching in by Ajv library.
Currently the whole schema is used as a cache key in each `ajv` instance[^1],
for very large schemas this leads to a lot of unnecessary memory usage since
a whole string representation of the schema is stored with each `compile()`.

The `ajv` cache is never used, since there's already a cache implementation
in the code base, but by passing a much shorter key to `ajv` we optimise
memory usage.

[^1]: https://ajv.js.org/guide/managing-schemas.html#cache-key-schema-vs-key-vs-id

### DIFF
--- a/src/middlewares/openapi.request.validator.ts
+++ b/src/middlewares/openapi.request.validator.ts
@@ -19,7 +19,7 @@ import { ParametersSchemaParser } from './parsers/schema.parse';
 import {
   ContentType,
   ajvErrorsToValidatorError,
-  augmentAjvErrors,
+  augmentAjvErrors, useAjvCache,
 } from './util';
 
 type OperationObject = OpenAPIV3.OperationObject;
@@ -28,7 +28,6 @@ type ReferenceObject = OpenAPIV3.ReferenceObject;
 type SecurityRequirementObject = OpenAPIV3.SecurityRequirementObject;
 type SecuritySchemeObject = OpenAPIV3.SecuritySchemeObject;
 type ApiKeySecurityScheme = OpenAPIV3.ApiKeySecurityScheme;
-
 export class RequestValidator {
   private middlewareCache: { [key: string]: RequestHandler } = {};
   private apiDoc: OpenAPIV3.DocumentV3 | OpenAPIV3.DocumentV3_1;
@@ -79,7 +78,7 @@ export class RequestValidator {
     const key = `${req.method}-${path}-${contentTypeKey}`;
 
     if (!this.middlewareCache[key]) {
-      const middleware = this.buildMiddleware(path, reqSchema, contentType);
+      const middleware = this.buildMiddleware(path, reqSchema, contentType, key);
       this.middlewareCache[key] = middleware;
     }
     return this.middlewareCache[key](req, res, next);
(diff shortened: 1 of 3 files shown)


cdimascio (approved):
Thanks. Appreciate this update! Looks good 👍

martinslota:
I'm noticing in our test suite that `express-openapi-validator` consumes more memory from version `5.5.0` onwards.

I have a feeling that this change might (ironically) be the cause of that: It turns out that despite the recommendation in `ajv`'s docs, [the `addSchema()` method](https://github.com/ajv-validator/ajv/blob/9050ba1359fb87cd7c143f3c79513ea7624ea443/lib/core.ts#L470-L492) stores the schema in `this.schemas` under the supplied key, but before it does that, it calls [the `_addSchema()` method](https://github.com/ajv-validator/ajv/blob/9050ba1359fb87cd7c143f3c79513ea7624ea443/lib/core.ts#L698-L727) which sticks the schema in `this._cache` as well. So, at the end of the day, the schema gets cached twice when using `addSchema()` instead of `compile()`.

martinslota:
> I'm noticing in our test suite that `express-openapi-validator` consumes more memory from version `5.5.0` onwards.
> 
> I have a feeling that this change might (ironically) be the cause of that: It turns out that despite the recommendation in `ajv`'s docs, [the `addSchema()` method](https://github.com/ajv-validator/ajv/blob/9050ba1359fb87cd7c143f3c79513ea7624ea443/lib/core.ts#L470-L492) stores the schema in `this.schemas` under the supplied key, but before it does that, it calls [the `_addSchema()` method](https://github.com/ajv-validator/ajv/blob/9050ba1359fb87cd7c143f3c79513ea7624ea443/lib/core.ts#L698-L727) which sticks the schema in `this._cache` as well. So, at the end of the day, the schema gets cached twice when using `addSchema()` instead of `compile()`.

A small follow-up on this: I have now taken version `5.5.0` and reverted the changes in this commit using [patch-package](https://www.npmjs.com/package/patch-package) and ran our tests again. The extra memory pressure disappeared.

0bex0:
> I'm noticing in our test suite that express-openapi-validator consumes more memory from version 5.5.0 onwards.
>
> I have a feeling that this change might (ironically) be the cause of that: It turns out that despite the recommendation in ajv's docs, [the addSchema() method](https://github.com/ajv-validator/ajv/blob/9050ba1359fb87cd7c143f3c79513ea7624ea443/lib/core.ts#L470-L492) stores the schema in this.schemas under the supplied key, but before it does that, it calls [the _addSchema() method](https://github.com/ajv-validator/ajv/blob/9050ba1359fb87cd7c143f3c79513ea7624ea443/lib/core.ts#L698-L727) which sticks the schema in this._cache as well. So, at the end of the day, the schema gets cached twice when using addSchema() instead of compile().

I believe you're right, and `_addSchema` continues to cache the key using the schema object itself meaning the memory issue w.r.t the cache itself is not addressed either. I will create a PR to revert this change both the redundant caching and object cache key need to be addressed in the `ajv` library itself.

PR #1070, opened 2025-05-20
Revert "Optimise Ajv cache usage"

0bex0:
Reverts cdimascio/express-openapi-validator#1062

[Comment](https://github.com/cdimascio/express-openapi-validator/pull/1062#issuecomment-2893209447) highlighted issues with the `ajv` library which make this fix redundant, while inadvertently adding extra memory usage by using a secondary cache which is not used.

### DIFF
--- a/src/middlewares/openapi.request.validator.ts
+++ b/src/middlewares/openapi.request.validator.ts
@@ -20,7 +20,6 @@ import {
   ContentType,
   ajvErrorsToValidatorError,
   augmentAjvErrors,
-  useAjvCache,
 } from './util';
 
 type OperationObject = OpenAPIV3.OperationObject;
@@ -29,6 +28,7 @@ type ReferenceObject = OpenAPIV3.ReferenceObject;
 type SecurityRequirementObject = OpenAPIV3.SecurityRequirementObject;
 type SecuritySchemeObject = OpenAPIV3.SecuritySchemeObject;
 type ApiKeySecurityScheme = OpenAPIV3.ApiKeySecurityScheme;
+
 export class RequestValidator {
   private middlewareCache: { [key: string]: RequestHandler } = {};
   private apiDoc: OpenAPIV3.DocumentV3 | OpenAPIV3.DocumentV3_1;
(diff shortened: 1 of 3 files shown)


cdimascio:
@0bex0 curious whether you've created an issue with ajv regarding the double caching. Would love to solve it there, upgrade the dep, and retain your fix (assuming it's feasible within a quick timeframe)

cdimascio:
@0bex0 curious whether you've created an issue with ajv regarding the double caching. Would love to solve it there, upgrade the dep, and retain your fix (assuming it's feasible within a quick timeframe)

Alternatively, we could maintain our own cache, essentially compile then cache the schema via a map we own rather than using ajv's addSchema

0bex0:
> Alternatively, we might maintain our own cache, essentially compile the schema on first use, then cache the schema via a map owned by the validator, rather than using ajv's addSchema.
>
> Essentially, this approach https://ajv.js.org/guide/managing-schemas.html#caching-schemas-in-your-code

The `ajv.compileAsync` call still caches the full schema object and `SchemaEnv` object within the `ajv` instance. The whole point of this internal cache is to avoid the re-compilation of schemas, but it's dependent on the exact same object reference for a schema to be used for the cache to actually work, which does not work with the `express-openapi-validator` implementation, since the same object reference is never used for any two schemas due to the logic in `Validator._schemaBody`.

I'm not sure (without any updates to the `ajv` library) there's a simple way to improve the memory usage w.r.t to the caching in this library at the moment. I'm going to raise an issue in the `ajv` library to at least fix the double caching and then this fix is retained, that would at least abate the problem of using the whole schema object as a cache key.

0bex0:
I have added an issue here: https://github.com/ajv-validator/ajv/issues/2554#issue-3079986159 and will address it in due course. This may not be possible until next week due to other commitments, but I have a fix in mind, yet to test it.

0bex0:
In order to be backwards compatible, the double caching logic has to stay in place in the `Ajv` library, however [I proposed adding an instance option](https://github.com/ajv-validator/ajv/pull/2555) which would allow users more flexibility w.r.t to how they choose to cache `SchemaEnv` objects.

PR #1101, opened 2025-07-23
add onError to validateRequestOptions

srsudar:
I would like to be able to slowly and safely introduce this to an existing project. It's possible that we have some misbehaving clients, an I'd like to be able to identify cases where we aren't correctly describing types, without necessarily breaking the clients.

It seems like to do this, and be aware of bad requests, I want an `onError` option in `ValidateRequestOptions`, the same way it exists in `ValidateResponseOptions`.

This PR is an attempt to accomplish that. I cribbed pretty liberally from the same tests for response validation, and I'd believe of course that I'm missing something. I welcome comments!

### DIFF
--- a/src/framework/ajv/options.ts
+++ b/src/framework/ajv/options.ts
@@ -30,14 +30,19 @@ export class AjvOptions {
   }
 
   get request(): RequestValidatorOptions {
-    const { allErrors, allowUnknownQueryParameters, coerceTypes, removeAdditional } = <
-      ValidateRequestOpts
-    >this.options.validateRequests;
+    const {
+      allErrors,
+      allowUnknownQueryParameters,
+      coerceTypes,
+      removeAdditional,
+      onError,
+    } = <ValidateRequestOpts>this.options.validateRequests;
     return {
       ...this.baseOptions(),
       allErrors,
       allowUnknownQueryParameters,
       coerceTypes,
+      onError,
       removeAdditional,
     };
   }
@@ -47,13 +52,8 @@ export class AjvOptions {
   }
 
   private baseOptions(): Options {
-    const {
-      coerceTypes,
-      formats,
-      validateFormats,
-      serDes,
-      ajvFormats,
-    } = this.options;
+    const { coerceTypes, formats, validateFormats, serDes, ajvFormats } =
+      this.options;
     const serDesMap = {};
     for (const serDesObject of serDes) {
       if (!serDesMap[serDesObject.format]) {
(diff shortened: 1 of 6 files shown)


srsudar:
@cdimascio -- any thoughts on this? I'd be happy to try some cleanup if it seems like something you're interested in merging. If not, I might maintain a fork. I think I'll need this to be able to safely begin integrating `express-openapi-validator` into a prod app.

src/framework/ajv/options.ts
```diff
@@ -30,14 +30,19 @@ export class AjvOptions {
...
-    const { allErrors, allowUnknownQueryParameters, coerceTypes, removeAdditional } = <
-      ValidateRequestOpts
-    >this.options.validateRequests;
+    const {
+      allErrors,
+      allowUnknownQueryParameters,
+      coerceTypes,
+      removeAdditional,
+      onError,
+    } = <ValidateRequestOpts>this.options.validateRequests;
```
cdimascio, line 39: Is this change necessary? Does the list include the complete set of user specified options? If not, it will change/ break existing behavior

src/framework/types.ts
```diff
@@ -61,6 +61,7 @@ export type ValidateRequestOpts = {
   allowUnknownQueryParameters?: boolean;
   coerceTypes?: boolean | 'array';
   removeAdditional?: boolean | 'all' | 'failing';
+  onError?: (err: InternalServerError, req: Request) => void;
```
cdimascio, line 64: Is the intend of this change to introduce onError to requests, mirroring what exists currently (but only) for responses?

srsudar: Yes that's right. My thinking is that we currently have three platforms hitting our APIs. I believe I've captured the the real-world usage of our APIs, but there are enough edge cases that I'd believe I've missed something. I'd like to be able to introduce validation slowly, basically making sure that we are correctly extracting the openapi interface before I begin actually serving 400s using the validator.

src/middlewares/openapi.request.validator.ts
```diff
@@ -218,7 +227,12 @@ export class RequestValidator {
           message: message,
         });
         error.errors = err.errors;
-        throw error;
+        if (this.requestOpts.onError) {
+          this.requestOpts.onError(error, req);
+          next();
```
cdimascio, line 232: do we need to call next() here? we should expect onError handlers to always throw, correct? is the intent to allow clients to eat errors?

PR #1135, opened 2026-02-12
Fix: Return error from attempted security scheme instead of first defined scheme

### LINKED ISSUE #1118: Error handling response multiple security strategies
**Describe the bug**
A service defines two security schemes: one bearer-based scheme and one apiKey-based scheme. Both have individual security handlers with their own internal validation logic. When a request uses the ApiKeyAuth scheme and its handler throws an error, the framework incorrectly returns the error from the bearer scheme (“missing token”), even though this scheme is not used for the request.

**To Reproduce**
A request is sent using the ApiKeyAuth header.
If the ApiKeyAuth handler fails, the middleware still reports the KeyCloakAuth error (“missing token”), instead of returning the actual error from the handler that was triggered.

**Actual behavior** 
The middleware always returns the missing-bearer-token error, regardless of which security scheme matches the request.

**Expected behavior**
The request should first be matched against the security scheme actually used. Only the handler for that scheme should execute. If this handler fails, its own error must be returned without falling back to other schemes.

**Examples and context**

The issue originates in the incorrect error-accumulation logic here:
https://github.com/cdimascio/express-openapi-validator/blob/master/src/middlewares/openapi.security.ts#L90

`
securitySchemes:
    KeyCloakAuth:
      type: http
      scheme: bearer
      bearerFormat: JWT
    ApiKeyAuth:
      type: apiKey
      in: header
      name: sessionid
`

aleksandar-kovacic:
## Problem
When multiple security schemes are defined (e.g., `BearerAuth OR ApiKeyAuth`) and a user provides credentials for only one scheme, the error returned was always from the first defined security scheme rather than the one actually attempted.

### Example
Given this OpenAPI security configuration:
```yaml
security:
  - BearerAuth: []
  - ApiKeyAuth: []
```

#### Before this fix
User sends request with an invalid ApiKey but no Bearer token.
ApiKey validation runs and fails with "Invalid API key".
❌ Error returned: "Authorization header required" (from BearerAuth, which wasn't attempted).

#### After this fix
User sends request with an invalid ApiKey but no Bearer token.
ApiKey validation runs and fails with "Invalid API key"
✅ Error returned: "Invalid API key provided" (from ApiKeyAuth, which was actually attempted)

## Solution
Added an attempted flag to track which security schemes received credentials. The error selection logic now prioritizes errors from schemes where authentication was actually attempted over schemes where credentials were missing.

Closes #1118

### DIFF
--- a/src/middlewares/openapi.security.ts
+++ b/src/middlewares/openapi.security.ts
@@ -22,6 +22,7 @@ interface SecurityHandlerResult {
   success: boolean;
   status?: number;
   error?: string;
+  attempted?: boolean; // true if credentials were provided and handler was called
 }
 
 function extractErrorsFromResults(results: (SecurityHandlerResult | SecurityHandlerResult[])[]) {
@@ -87,7 +88,10 @@ export function security(
         next();
       } else {
         const errors = extractErrorsFromResults(results);
-        throw errors[0];
+        // Prioritize errors where authentication was actually attempted
+        const attemptedErrors = errors.filter(e => e.attempted);
+        const errorToThrow = attemptedErrors.length > 0 ? attemptedErrors[0] : errors[0];
+        throw errorToThrow;
       }
     } catch (e) {
       const message = e?.error?.message || 'unauthorized';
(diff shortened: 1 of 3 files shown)


Fridious:
@cdimascio I have made all the necessary changes. Is there anything else to do?

Gatos90:
Pls merge this. Im having the same problem!

cdimascio (approved):

PR #1197, opened 2026-05-27
Move @types/multer to devDependencies

mdmower-csnw:
Move `@types/multer` to dev dependencies so that applications shipping prod-only dependencies do not need this package and all of its `@types/*` transitive dependencies.


mdmower-csnw:
@cdimascio - Would you be willing to review this?

